PR #3, opened 2022-08-24
Initial gRPC

chromevoid:


pkg/function/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return mappedDatumList.GetElements(), nil
+}
+
+// TODO: use a channel to accept datumStream?
+
+func (c *Client) ReduceFn(ctx context.Context, datumStream []*functionpb.Datum) ([]*functionpb.Datum, error) {
+	stream, err := c.grpcClt.ReduceFn(ctx)
+	if err != nil {
+		log.Fatalf("failed to execute c.grpcClt.ReduceFn(): %v", err)
```
whynowy, line 71: Return err instead fatal.

pkg/function/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	for _, datum := range datumStream {
+		if err := stream.Send(datum); err != nil {
+			log.Fatalf("failed to execute stream.Send(%v): %v", datum, err)
```
whynowy, line 75: same here.

pkg/function/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+	reducedDatumList, err := stream.CloseAndRecv()
+	if err != nil {
+		log.Fatalf("failed to execute stream.CloseAndRecv(): %v", err)
```
whynowy, line 80: and this.

pkg/function/configs.go
```diff
@@ -0,0 +1,6 @@
+package function
+
+const (
+	Protocol = "unix"
+	Addr     = "/var/run/numaflow/grpc.sock"
```
whynowy, line 5: `function.sock`?

pkg/function/client/options.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1/funcmock"
+)
+
+type options struct {
+	mockClient gRPClientOption
+}
+
+type Option interface {
```
whynowy, line 11: Use the approach like https://github.com/numaproj/numaflow/blob/main/pkg/shared/clients/jetstream/options.go#L23, which is simpler.

pkg/function/example/forward_msg/server/main.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/numaproj/numaflow-go/pkg/function/server"
+)
+
+func handle(ctx context.Context, key string, msg []byte) (functionsdk.Messages, error) {
+	// directly forward the input to the output
+	return functionsdk.MessagesBuilder().Append(functionsdk.MessageTo(key, msg)), nil
+}
+
+func main() {
+	server.NewServer().RegisterMapper(functionsdk.DoFunc(handle)).Start()
```
whynowy, line 16: `server.New().Regis...`

No need to have `Server` in the function name.

pkg/function/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+func (c *Client) DoFn(ctx context.Context, datum *functionpb.Datum) ([]*functionpb.Datum, error) {
+	mappedDatumList, err := c.grpcClt.DoFn(ctx, datum)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute c.grpcClt.DoFn(): %w", err)
+	}
+
+	return mappedDatumList.GetElements(), nil
+}
+
+// TODO: use a channel to accept datumStream?
```
whynowy, line 66: this should be a channel.

pkg/function/client/options.go
chromevoid: @vigith suggested https://github.com/uber-go/guide/blob/master/style.md#functional-options
I'm ok with either one

vigith: The reason I asked for struct type here is so that we can attach other complex functions to the data type later. These are not the usual options, but one can imagine these like `GetXXX` and `SetXXX`.

vigith: I am thinking of using `opts ...MapOptions` available in the `handle` function. Eg., one could do `opts.GetWatermark()`.

.gitignore
```diff
@@ -13,3 +13,7 @@ test/*.cov
 site/
 /go-diagrams/
 *.idea/
+
+# client code for local testing only
```
vigith, line 17: do we need to commit this to .gitignore?

pkg/function/client/options.go
vigith: looks like these are `options` to gprc. This can be implemented any way we like. My comments are mostly for `pkg/function/options.go`

pkg/function/service.go
```diff
@@ -0,0 +1,49 @@
...
+	messages, err := fs.Mapper.HandleDo(ctx, d.GetKey(), d.GetValue())
+	if err != nil {
+		return nil, err
+	}
+	var elements []*functionpb.Datum
+	for _, m := range messages.Items() {
+		elements = append(elements, &functionpb.Datum{
+			Key:            m.Key,
+			Value:          m.Value,
+			EventTime:      d.GetEventTime(), // should provide a way for the user to update the event time
```
vigith, line 34:
```suggestion
			EventTime:      d.GetEventTime(),
```

we don't have to make the user update the event time, perhaps in the source vertex we can also ask the UDF to return eventtime also. we cannot change the eventtime after ingestion.

.gitignore
chromevoid: I have this file on my local for testing purpose... 
and I want to keep it.......

pkg/function/client/options.go
whynowy: How to make it happen in the `handle` function? I could imagine the function looks like `handle(key string, value []byte, meta metadata)`, is there a way to do it like `handle(key string, value []byte, opts ...Options)`, where `opts` is an array of interface?

chromevoid: @whynowy I have this experimental code for handler function options
https://github.com/numaproj/numaflow-go/tree/grpc-option

The idea is to first set up the MapOption when registering the mapper.
Then, in the DoFn, the MapOption is converted to a HandleOption which is passed to the HandleDo().

whynowy: > @whynowy I have this experimental code for handler function options https://github.com/numaproj/numaflow-go/tree/grpc-option
> 
> The idea is to first set up the MapOption when registering the mapper. Then, in the DoFn, the MapOption is converted to a HandleOption which is passed to the HandleDo().

I understand that. 

The developer/user should not be responsible for populate those options in their code, instead, what should be done is, the SDK populate them somewhere (e.g. a struct object), and the developer is only supposed to call get functions.

whynowy: The following part of code should be not written by each of the SDK users.

```
options := functionsdk.MapOptions{
		EventTime: time.Time{},
	}
	for _, o := range inputOptions {
		o.Get(&options)
	}
```

chromevoid: I've updated the func sdk client option implementation

pkg/function/client/client.go
```diff
@@ -0,0 +1,88 @@
...
+// Client contains the grpc connection and the grpc client.
+type Client struct {
+	conn    *grpc.ClientConn
+	grpcClt functionpb.UserDefinedFunctionClient
+}
+
+// New creates a new client object.
+func New(inputOptions ...Option) (*Client, error) {
+	var opts = &options{
+		mockClient: nil,
```
whynowy, line 23: This is an interesting approach to get mock integrated, do you have an example that any other projects are doing this way? The brings some concern to me, that a mockclient becomes an official option of calling this API, which is supposed to be an internal stuff for testing.

pkg/function/client/options.go
whynowy: Another approach specifically for Golang is, get those stuff into `context`, for example, we have a struct to wrap all of those options, instead of putting the struct in the function signature, put it in the context.

For Python, I guess the common practice is having a `metadata` in the function signature.

pkg/function/client/client.go
chromevoid: > do you have an example that any other projects are doing this way?

No.. it's just my ideas....

The numaflow imports the client and needs to have a way to get a mock client for testing. 
client.New() is called here for real use case
https://github.com/chromevoid/numaflow/blob/udf-grpc-client/pkg/udf/applier/uds_grpc.go#L23-L29
and here for tests
https://github.com/chromevoid/numaflow/blob/udf-grpc-client/pkg/udf/applier/uds_grpc_test.go#L22-L25

.gitignore
vigith: you can override using your global gitignore

pkg/function/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+// Start starts the gRPC server via unix domain socket at configs.Addr.
+func (s *server) Start(inputOptions ...Option) {
+	cleanup := func() {
+		if _, err := os.Stat(function.Addr); err == nil {
+			if err := os.RemoveAll(function.Addr); err != nil {
+				log.Fatal(err)
+			}
+		}
+	}
+	cleanup()
```
whynowy, line 45: Move to line 54 with `cleanup(opts.sockAddr)`

pkg/function/client/client.go
vigith: we shouldn't leak mocking to the official option.

whynowy (approved):

pkg/function/server/server_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
+	functionsdk "github.com/numaproj/numaflow-go/pkg/function"
+	"github.com/numaproj/numaflow-go/pkg/function/client"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_server_Start(t *testing.T) {
+	testSock := "/tmp/uds/numaflow.sock"
```
vigith, line 15: can we create a tmp file and delete after testing?
```
file, err := ioutil.TempFile(...)
assert.NoError(err)
defer os.Remove(file.Name())
```

vigith (approved):

PR #5, opened 2022-08-27
feat: start both http and grpc server in udf

chromevoid:
Signed-off-by: jyu6 <juanlu_yu@intuit.com>

start both http and grpc server in udf

### DIFF
--- a/pkg/function/server/server.go
+++ b/pkg/function/server/server.go
@@ -1,10 +1,12 @@
 package server
 
 import (
+	"context"
 	"log"
 	"net"
 	"os"
 
+	httpfuncsdk "github.com/numaproj/numaflow-go/function"
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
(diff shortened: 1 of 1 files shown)


pkg/function/server/server.go
```diff
@@ -62,4 +64,19 @@ func (s *server) Start(inputOptions ...Option) {
 	if err := grpcSvr.Serve(lis); err != nil {
 		log.Fatalf("failed to start the gRPC server: %v", err)
 	}
+
+	httpfuncsdk.Start(context.Background(), func(ctx context.Context, key, msg []byte) (httpfuncsdk.Messages, error) {
```
whynowy, line 68: It can not be executed when gRPC server is started, use a goroutine to start it before starting the gRPC server.

vigith:
I am thinking we should remove the support for HTTP. In golang it is easy to do a go routine and serve both HTTP and gRPC requests. In python, it won't be that easy. Maybe we should enforce `namespaced` installation starting with the introduction of gRPC. 

WDYT?

PR #11, opened 2022-08-30
chore: refactor map and reduce

### LINKED ISSUE #9: refactor map and reduce interface
Let's refactor Map and Reduce handle as follows. This can give richer semantics on how to get other metadata from Datum. 
Please note that `Datum` is an interface that supports
* `Key`
* `Value`
* `Eventtime`
* `Watermark`

Please note that `Reduce` handle has a `metadata` arg which has static information for the entire reduce context.

# Map Handle

```golang
func map_handle(ctx context.Context, key string, d Datum) (functionsdk.Messages) {
	// directly forward the input to the output
	val := d.Value()
	eventTime := d.Eventtime()
	watermark := d.Watermark()

	var resultKey string = key
	var resultVal []byte = val
	return functionsdk.MessagesBuilder().Append(functionsdk.MessageTo(resultKey, resultVal)), nil
}
```

# Reduce Handle

```golang
func reduce_handle(ctx context.Context, key string, reduceCh <-chan Datum, md Metadata) (functionsdk.Messages) {
	intervalWindow := md.IntervalWindow()
	var resultKey string = key
	var resultVal []byte
	// directly forward the input to the output
	for d := range reduceCh {
		val := d.Value()
		eventTime := d.Eventtime()
		watermark := d.Watermark()
	}
	return functionsdk.MessagesBuilder().Append(functionsdk.MessageTo(resultKey, resultVal)), nil
}
```

chromevoid:
Signed-off-by: jyu6 <juanlu_yu@intuit.com>

fixes #9


chromevoid:
I can revert the last commit if it's not good...

vigith:
after thinking, i think we need to pass `key string` to the func and remove `Key()` from the interface.

chromevoid:
Please check the new todos I added in the last commit [f70abcb](https://github.com/numaproj/numaflow-go/pull/11/commits/f70abcb5227851ea57a59c81628df53410b3cce2) .....

pkg/function/service.go
```diff
@@ -83,7 +83,7 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
 	//   we can use grpc metadata instead? for example:
 	if grpcMD, ok := metadata.FromIncomingContext(stream.Context()); ok {
 		// get Key
-		keyValue := grpcMD.Get("key")
+		keyValue := grpcMD.Get(DatumKey)
 		if len(keyValue) != 1 {
 			// TODO: or if there's multiple values we directly use the first one?
```
chromevoid, line 88: or actually we are the only one to implement the client... we can directly use `grpcMD.Get(DatumKey)[0]` without any checks?......

pkg/function/service.go
```diff
@@ -20,9 +45,14 @@ func (fs *Service) IsReady(context.Context, *emptypb.Empty) (*functionpb.ReadyRe
 	return &functionpb.ReadyResponse{Ready: true}, nil
 }
 
-// DoFn applies a function to each datum element
-func (fs *Service) DoFn(ctx context.Context, d *functionpb.Datum) (*functionpb.DatumList, error) {
-	messages, err := fs.Mapper.HandleDo(ctx, d.GetKey(), d.GetValue())
+// MapFn applies a function to each datum element
+func (fs *Service) MapFn(ctx context.Context, d *functionpb.Datum) (*functionpb.DatumList, error) {
+	var hd = handlerDatum{
+		value:     d.GetValue(),
```
vigith, line 51: to make it consistent, let's also pass key in the Map also via the gRPC Metatata.

pkg/function/example/sum/main.go
```diff
@@ -0,0 +1,38 @@
...
+	for d := range reduceCh {
+		val := d.Value()
+		eventTime := d.EventTime()
+		_ = eventTime
+		watermark := d.Watermark()
+		_ = watermark
+
+		v, err := strconv.Atoi(string(val))
+		if err != nil {
+			continue
```
vigith, line 28:
```suggestion
			fmt.Println("err, ", err)
			continue
```

pkg/function/service.go
```diff
@@ -20,9 +45,26 @@ func (fs *Service) IsReady(context.Context, *emptypb.Empty) (*functionpb.ReadyRe
...
-	messages, err := fs.Mapper.HandleDo(ctx, d.GetKey(), d.GetValue())
+// MapFn applies a function to each datum element
+func (fs *Service) MapFn(ctx context.Context, d *functionpb.Datum) (*functionpb.DatumList, error) {
+	var key string
+
+	// get key from gPRC metadata
+	if grpcMD, ok := metadata.FromIncomingContext(ctx); ok {
+		keyValue := grpcMD.Get(DatumKey)
+		if len(keyValue) > 1 {
+			return nil, fmt.Errorf("expect one key but got multiple keys")
```
vigith, line 56: let's do a `%d` on len of the keys and perhaps even list it via `%v` ? will make debugging easier.

pkg/function/service.go
```diff
@@ -20,9 +45,26 @@ func (fs *Service) IsReady(context.Context, *emptypb.Empty) (*functionpb.ReadyRe
...
+		} else if len(keyValue) == 0 {
```
vigith, line 57: it is okay for the key to be missing. if key is missing, make it `""` ?

pkg/function/service.go
```diff
@@ -41,9 +84,65 @@ func (fs *Service) DoFn(ctx context.Context, d *functionpb.Datum) (*functionpb.D
...
+		ctx      = stream.Context()
+		key      string
+		reduceCh = make(chan Datum)
+		md       Metadata
+	)
+
+	// get key and metadata from gPRC metadata
+	if grpcMD, ok := metadata.FromIncomingContext(ctx); ok {
+		// get Key
+		keyValue := grpcMD.Get(DatumKey)
```
vigith, line 98: same comment as of map

pkg/function/service.go
```diff
@@ -41,9 +84,65 @@ func (fs *Service) DoFn(ctx context.Context, d *functionpb.Datum) (*functionpb.D
...
+		datumList []*functionpb.Datum
+		wg        sync.WaitGroup
+	)
+
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+		messages, err := fs.Reducer.HandleDo(ctx, key, reduceCh, md)
+		if err != nil {
+			// TODO: deal with err
```
vigith, line 118: at least print it, let's not swallow

pkg/function/service.go
```diff
@@ -41,9 +84,65 @@ func (fs *Service) DoFn(ctx context.Context, d *functionpb.Datum) (*functionpb.D
...
+				Value:     msg.Value,
+				EventTime: &functionpb.EventTime{EventTime: timestamppb.New(time.Time{})}, // TODO: what's the correct value?...
+				Watermark: &functionpb.Watermark{Watermark: timestamppb.New(time.Time{})},
+			})
+		}
+	}()
+
+	for {
+		datum, err := stream.Recv()
+		if err == io.EOF {
```
vigith, line 134: we are swallowing other errors here

vigith (commented):
do we need to think about error handling?

Makefile
```diff
@@ -1,6 +1,8 @@
+.PHONY: all
+all: proto generate test
 
 .PHONY: test
-test: proto generate
+test:
```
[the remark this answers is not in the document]
chromevoid: I ran "make test" a lot more than the other two commands.
I added "make all" for "proto generate test" purpose..

pkg/function/interface.go
```diff
@@ -2,33 +2,57 @@ package function
...
 }
 
 // MapHandler is the interface of map function implementation.
 type MapHandler interface {
 	// HandleDo is the function to process each coming message
-	HandleDo(ctx context.Context, key string, msg []byte) (Messages, error)
+	HandleDo(ctx context.Context, key string, datum Datum) (Messages, error)
 }
 
 // ReduceHandler is the interface of reduce function implementation.
```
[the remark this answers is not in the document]
chromevoid: on a second thought.. is it still a TODO?....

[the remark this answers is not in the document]
vigith: looks complete to me 😄

chromevoid:
> do we need to think about error handling?

we can do something like a customized error handler function as an option

```go
// WithCustomAutoRefreshingErrHandler defines how to handle errs from autoRefreshing.
func WithCustomErrHandler(errHandler func(err error)) Option {
	return func(opts *cacheOptions) {
		opts.errHandler = errHandler
	}
}
```
and a default error handler could be
```go
func defaultAutoRefreshingErrHandler(err error) {
	log.Error(err)
}
```

pkg/function/service.go
```diff
@@ -2,11 +2,36 @@ package function
...
+	"time"
 
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// handlerDatum implements the Datum interface and is used in the map and reduce handlers.
+type handlerDatum struct { // TODO: better name??...
```
vigith, line 17:
```suggestion
type handlerDatum struct {
```

vigith (commented):
let's not send back watermark and eventtime

vigith (approved):

pkg/function/interface.go
```diff
@@ -2,33 +2,61 @@ package function
...
+type IntervalWindow interface {
+	StartTime() time.Time
+	EndTime() time.Time
 }
 
 // MapHandler is the interface of map function implementation.
 type MapHandler interface {
 	// HandleDo is the function to process each coming message
-	HandleDo(ctx context.Context, key string, msg []byte) (Messages, error)
+	HandleDo(ctx context.Context, key string, datum Datum) (Messages, error)
```
vigith, line 40: let's not support `error` from Map handle

pkg/function/interface.go
```diff
@@ -2,33 +2,61 @@ package function
...
 }
 
 // ReduceHandler is the interface of reduce function implementation.
 type ReduceHandler interface {
-	// TODO
+	HandleDo(ctx context.Context, key string, reduceCh <-chan Datum, md Metadata) (Messages, error)
```
vigith, line 45: let's not support `error` from reduce handle

vigith (approved):

PR #16, opened 2022-09-15
feature: udsink gRPC

chromevoid:
support udsink gRPC

Tested using the `pkg/sink/examples/udsink_log.go` example
image


pkg/sink/configs.go
```diff
@@ -0,0 +1,6 @@
+package sink
+
+const (
+	Protocol = "unix"
+	Addr     = "/var/run/numaflow/function.sock"
+)
```
chromevoid, line 6: I first refactored the configs to be pkg/configs
then when I was working on the numaflow project I found that I need to import one more pkg/configs
so I reverted the refactor and added a separate configs under sink

let me know what do you think

pkg/sink/types.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Message is used to wrap the message written to the user defined sink
+// TODO: The numaflow project use sinksdk.Message as the input argument for Apply function.
+//
+//	Therefore, since the Message type doesn't have event time and watermark, these two fields
+//	won't be populated into the Datum.event_time and Datum.watermark.
+//	In short:
+//	  Add EventTime and Watermark in the Message type
+//	  or
+//	  Delete event_time and watermark in the sink proto file.
```
chromevoid, line 11: Please check this TODO

chromevoid: I feel like we only need Datum...
Message type is only used in the numaflow project. We can replace [sinksdk.Message](https://github.com/numaproj/numaflow/pull/174/files#diff-9bee584308ea77899e2541fc380d19957f6bde1e5f582d960d66609516f28556R46) with sinkpb.Datum.

chromevoid: removed message type..

chromevoid:
Do we need key for sink.Datum type???

vigith (approved):

PR #18, opened 2022-09-20
feature: mock code for reduce client

yhl25:
Signed-off-by: yhl01 <Yashash_HL@intuit.com>

need this to test reduce in numaflow


pkg/apis/proto/function/v1/mockgen.go
```diff
@@ -1,3 +1,4 @@
 package v1
 
 //go:generate mockgen -destination funcmock/funcmock.go -package funcmock github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1 UserDefinedFunctionClient
+//go:generate mockgen -destination funcmock/reducemock.go -package funcmock github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1 UserDefinedFunction_ReduceFnClient
```
chromevoid, line 4: We may use one line for both client and stream mock.

chromevoid: What do others think?..

yhl25: I didn't know we could do that.
yeah, it makes sense to keep them in one single file.

chromevoid (approved):

vigith (approved):

PR #22, opened 2022-11-10
passing metadata to reduce handler

yhl25:
Fixes https://github.com/numaproj/numaflow/issues/334

Signed-off-by: Yashash H L <yashash_hl@intuit.com>

### DIFF
--- a/pkg/function/service.go
+++ b/pkg/function/service.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strconv"
 	"sync"
 	"time"
 
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
-	"google.golang.org/grpc/metadata"
+	grpcmd "google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/function/service.go
```diff
@@ -81,26 +117,47 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.Datum) (*functionpb.
...
+	if err != nil {
+		return err
+	}
+
+	et, err = getValueForKey(grpcMD, WinEndTime)
+	if err != nil {
+		return err
 	}
 
+	startTime, _ = strconv.Atoi(st)
```
ashwinidulams, line 152: should use strconv.ParseInt functions here for int64 values.

pkg/function/service.go
```diff
@@ -81,26 +117,47 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.Datum) (*functionpb.
...
+	endTime, _ = strconv.Atoi(et)
+
+	// create interval window interface using the start and end time
+	iw := NewIntervalWindow(time.UnixMilli(int64(startTime)), time.UnixMilli(int64(endTime)))
+
+	// create metadata using interval window interface
+	md = NewMetadata(iw)
```
ashwinidulams, line 159: do we need Metadata? why dont we make it explicit and pass in IntervalWindow as an argument?
we dont expect any additional metadata here.

pkg/function/service.go
```diff
@@ -142,3 +199,21 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
...
 	})
 }
+
+func getValueForKey(md grpcmd.MD, key string) (string, error) {
+	var value string
+
+	keyValue := md.Get(key)
+
+	if len(keyValue) > 1 {
+		return value, fmt.Errorf("expected extactly one value for key %s in metadata but got %d values", key, len(keyValue))
```
ashwinidulams, line 209: log values as well.

pkg/function/service.go
yhl25: yes, done

pkg/function/service.go
yhl25: done

vigith (approved):
LTGM

pkg/function/service.go
chromevoid: I prefer Metadata. If we are to have more metadata fields in the future, we can directly add in the `md` and the user doesn't have to update their reduce handler signature. e.g. https://github.com/numaproj/numaflow-go/blob/main/pkg/function/examples/sum/main.go#L12

chromevoid (approved):

PR #24, opened 2022-12-09
chore: unify examples

chromevoid:
simple reduce counter example


vigith:
should we rename these example repos with the prefix `map-` or `reduce-`?

chromevoid:
> should we rename these example repos with the prefix `map-` or `reduce-`?

the docker image repo?

vigith:
> the docker image repo?
your call 😄

pkg/function/examples/counter/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+FROM alpine:3.12.3 as base
+RUN apk update && apk upgrade && \
+    apk add ca-certificates && \
+    apk --no-cache add tzdata
+
+COPY dist/go-counter-example /bin/go-counter-example
+RUN chmod +x /bin/go-counter-example
+
+####################################################################################################
+# evenodd
```
whynowy, line 13: Comment.

whynowy:
It would be good to put the example images into different groups, for example, `quay.io/numaio/udf/go-counter-example` - feel free to come up with better names.

vigith (approved):

PR #25, opened 2022-12-10
feat: udsink gRPC stream

chromevoid:
https://github.com/numaproj/numaflow/issues/335
Compare to https://github.com/numaproj/numaflow-go/pull/23, this way we don't need to update the client side (except for the sink test...)


chromevoid:
I don't update the client SinkFn to use a channel as the input because the way numaflow call this function is from `Write` to `Apply` to this `SinkFn`.
The `Write` has to implement the interface where it accepts an array of messages. Then, this array of messages will be converted to sinksdk's datum type and passed to `Apply`, and `Apply` calls `client.SinkFn`...

Since anyway the `Write` has to use an array (unlike reduce where we have a channel in the numaflow (correct?)), there's no need to update the client.SinkFn... (in my opinion)

wdyt?

chromevoid:
> I don't update the client SinkFn to use a channel as the input because the way numaflow calls this function is from `Write` to `Apply` to this `SinkFn`. The `Write` has to implement the interface where it accepts an array of messages. Then, this array of messages will be converted to sinksdk's datum type and passed to `Apply`, and `Apply` calls `client.SinkFn`...
> 
> Since anyway the `Write` has to use an array (unlike reduce where we have a channel in the numaflow (correct?)), there's no need to update the client.SinkFn... (in my opinion)
> 
> wdyt?


https://github.com/numaproj/numaflow/blob/main/pkg/sinks/udsink/sink.go#L96
=>
https://github.com/numaproj/numaflow/blob/a46fb9647c990ab69b0071bfc0ea6dfd6019f1bf/pkg/sinks/udsink/udsink_grpc.go#L62
=>
`responseList, err := u.client.SinkFn(ctx, dList)`

vigith (approved):
since our README is very version specific, we should move the README to godoc.

vigith:
can you add a `doc.go`

chromevoid:
image
image

vigith (approved):

PR #27, opened 2022-12-16
chore: add redis udsink to support numaflow E2E testing

KeranYang:
Tested with https://github.com/numaproj/numaflow/pull/424/files#diff-51c5d58a23c4f30ae52e8538c639e273c4ff64026fd1ae2992b2b8e6435b4a97


vigith:
This should be in https://github.com/numaproj/numaflow-sinks. You can create a top-level sink calling it something like `redis-e2e-test`.

This repo is just for SDK.

PR #30, opened 2023-01-10
feat: add sdk for implementing user defined source data transformer

KeranYang:


pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
+
+option go_package = "github.com/numaproj/numaflow-go/pkg/apis/proto/source/v1";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/empty.proto";
+
+package source.v1;
+
+service UserDefinedSourceTransformer {
```
whynowy, line 10: Do not use `UserDefinedSourcexxxx`, as we will have real `UserDefinedSource` in the future (not transformer). Probably just `SourceTransformer`?

pkg/function/client/client.go
```diff
@@ -18,18 +19,17 @@ type client struct {
 }
 
 // New creates a new client object.
-func New(inputOptions ...Option) (*client, error) {
-
-	var opts = &options{
-		sockAddr: function.Addr,
+func New(inputOptions ...sharedutils.Option) (*client, error) {
```
whynowy, line 22: Do not touch the `Option` in other packages, let them separated, even there's duplication.

pkg/source/client/client.go
```diff
@@ -0,0 +1,64 @@
+package client
```
whynowy, line 1: Use a different package other than `pkg/source/xxx`, that should be left to real user defined source.

pkg/function/client/client.go
KeranYang: The grpc client/server implementation is very similar across UDSink/UDFunction/UDTransformer. The `Options` struct is exactly same. Do we see customization for future UDxxx?

PR #31, opened 2023-01-12
feat: add mapTFn to allow user defined function to assign a new event time to message

KeranYang:
This feature is to be used by source vertex. The benefits are 1. user can assign a new event time to the message and 2. user can define filtering logics to filter out messages that don't need to be passed to source isb.

Signed-off-by: Keran Yang <yangkr920208@gmail.com>


pkg/function/server/server.go
```diff
@@ -26,35 +26,54 @@ func New() *server {
...
 func (s *server) RegisterMapper(m functionsdk.MapHandler) *server {
 	s.svc.Mapper = m
 	return s
 }
 
+// RegisterMapperT registers the mapT operation handler to the server.
+// Example:
+//
+//	func handle(ctx context.Context, key string, data functionsdk.Datum) functionsdk.MessageTs {
+//		_ = data.EventTime() // Event time is available
```
vigith, line 48: I am thinking how we will get `EventTime` and `Watermark` here if we are to use this new handler to get event-time and then assign watermark.

pkg/function/server/server_test.go
```diff
@@ -78,20 +79,71 @@ func Test_server_map(t *testing.T) {
...
+				}),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// note: using actual UDS connection
 
-	var testKey = "reduce_key"
+			go New().RegisterMapperT(tt.fields.mapTHandler).Start(context.Background(), WithSockAddr(file.Name()))
```
vigith, line 107: will this go-routine be leaked after the test is over?

pkg/function/service.go
```diff
@@ -114,6 +116,42 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.Datum) (*functionpb.
...
+			return nil, fmt.Errorf("expect extact one key but got %d keys", len(keyValue))
+		} else if len(keyValue) == 1 {
+			key = keyValue[0]
+		} else {
+			// do nothing: the length equals zero is valid, meaning the key is an empty string ""
+		}
+	}
+	var hd = handlerDatum{
+		value:     d.GetValue(),
+		eventTime: d.GetEventTime().EventTime.AsTime(),
```
vigith, line 137: I am thinking about how we will assign EventTime and Watermark here if we are to use this new user-defined handler to get event-time and then assign watermark.

pkg/function/server/server.go
KeranYang: My understanding is that within the input datum, although we already have EventTime and Watermark, whether or not they are valid, is up to the user to determine.

User can implement the UDF to completely ignore the passed in EventTime and Watermark and assign an irrelevant new EventTime. User can also choose to consider the passed in EventTime as valid and maybe generate a new EventTime based on the passed-in one(plus one minute if it makes sense).

At the end of the UDF execution, the only thing numaflow cares about is the `functionsdk.MessageTs` returned.

pkg/function/service.go
KeranYang: See comments above, I am thinking let user determine whether or not EventTime and Watermark assigned here is valid.

pkg/function/server/server_test.go
KeranYang: Good catch. Will fix. Thanks!

pkg/function/server/server.go
vigith: fair point, please document this.

pkg/function/server/server.go
```diff
@@ -105,5 +125,10 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
 	<-ctxWithSignal.Done()
 	log.Println("Got a signal: terminating gRPC server...")
 	defer log.Println("Successfully stopped the gRPC server")
-	grpcSvr.GracefulStop()
+	s.grpcServer.GracefulStop()
+}
+
+func (s *server) Stop() {
```
whynowy, line 131: Why do we need this?

pkg/sink/server/server.go
```diff
@@ -88,5 +90,10 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
 	<-ctxWithSignal.Done()
 	log.Println("Got a signal: terminating gRPC server...")
 	defer log.Println("Successfully stopped the gRPC server")
-	grpcSvr.GracefulStop()
+	s.grpcServer.GracefulStop()
+}
+
+func (s *server) Stop() {
```
whynowy, line 96: why do you need this?

pkg/function/server/server.go
KeranYang: I added this `Stop` function such that in our unit test, we can explicitly call this method to stop the grpc server after we finish sending all out test requests. (to address comment: https://github.com/numaproj/numaflow-go/pull/31#discussion_r1067655652)

whynowy: What you need to do is using a cancellable `ctx` for `Start()`, and then `cancel()` that ctx will shut down the server.

pkg/apis/proto/function/v1/udfunction.proto
```diff
@@ -8,10 +8,14 @@ import "google/protobuf/empty.proto";
 package function.v1;
 
 service UserDefinedFunction {
-  // Applies a function to each datum element.
+  // MapFn applies a function to each datum element without modifying event time.
   rpc MapFn(Datum) returns (DatumList);
 
-  // Applies a reduce function to a datum stream.
+  // MapTFn applies a function to each datum element.
```
vigith, line 14: Please document that this can be called from source transformer.

pkg/function/doc.go
```diff
@@ -25,6 +25,36 @@
     server.New().RegisterMapper(functionsdk.MapFunc(handle)).Start(context.Background())
   }
 */
+//
+// Example MapT (extracting event time from the datum payload)
+// MapT includes both Map and EventTime assignment functionalities.
```
vigith, line 30: please call out this can be used only at source.

pkg/function/examples/assign_event_time/Dockerfile
```diff
@@ -0,0 +1,20 @@
+####################################################################################################
+# base
+####################################################################################################
+FROM alpine:3.12.3 as base
```
vigith, line 4: alpine:3.17

vigith (approved):

PR #33, opened 2023-01-20
Add a user defined source data transformer example for numaflow e2e test

KeranYang:
# Event Time Filter

This is a simple User Defined Function example which receives a message, applies the following data transformation, and returns the message.

## Data Transformation
* If the message event time is **before** year 2022, drop the message.
* If it's **within** year 2022, update the key to "within_year_2022" and update the message event time to Jan 1st 2022.
* Otherwise(exclusively after year 2022), update the key to "after_year_2022" and update the message event time to Jan 1st 2023.

Signed-off-by: Keran Yang <yangkr920208@gmail.com>

### DIFF
--- a/pkg/function/examples/event_time_filter/Makefile
+++ b/pkg/function/examples/event_time_filter/Makefile
@@ -4,7 +4,7 @@ build:
 
 .PHONY: image
 image: build
-	docker build -t "quay.io/numaio/numaflow-go/event-time-filter:v1" --target event-time-filter .
+	docker build -t "quay.io/numaio/numaflow-go/mapt-event-time-filter::latest" --target event-time-filter .
 
 clean:
 	-rm -rf ./dist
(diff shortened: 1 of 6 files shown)


vigith:
Please make sure for images, the image name starts with `mapT`; we have a convention for the example image names, https://quay.io/organization/numaio

pkg/function/examples/event_time_filter/Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: build
+build:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -o ./dist/event-time-filter-example main.go
+
+.PHONY: image
+image: build
+	docker build -t "quay.io/numaio/numaflow-go/mapt-event-time-filter::latest" --target event-time-filter .
```
vigith, line 7:
```suggestion
	docker build -t "quay.io/numaio/numaflow-go/mapt-event-time-filter:latest" --target event-time-filter .
```

I have never seen `::` syntax in image tag, is it valid?

vigith (approved):

pkg/function/examples/event_time_filter/Makefile
KeranYang: oops. typo

PR #34, opened 2023-01-25
feat: improved performance using one grpc stream per window

yhl25:


vigith (approved):

pkg/function/service.go
```diff
@@ -230,15 +215,42 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
...
+			chanMap[d.Key] = ch
+
+			wg.Add(1)
+			go func(key string, ch chan Datum) {
+				defer wg.Done()
+				messages := fs.Reducer.HandleDo(ctx, key, ch, md)
+				mu.Lock()
+				defer mu.Unlock()
+				for _, msg := range messages {
+					datumList = append(datumList, &functionpb.Datum{
```
whynowy, line 231: Instead of using a lock, would it be better to use a channel for the returned datum?

vigith: We need to rethink the return https://github.com/numaproj/numaflow/issues/509

PR #36, opened 2023-02-12
refactor: return error for server start

### LINKED ISSUE #35: server Start to return error instead of calling log.Fatal()
In golang app, usually we handle the error at top level command. Sub function calls just return the error to the upper function for it to decide what to do with the error.
 
Can Numaflow UDF grpc server start method return the error instead of calling log.FatalF for error handling?

Example usage with Cobra

```go
var myCmd = &cobra.Command{
	Use:          "my-cmd",
	RunE:         runQuery,
}

func init() {
	RootCmd.AddCommand(myCmd)
}

func handle(ctx context.Context, key string, d fn.Datum) fn.Messages {
	msg := d.Value()
	// If msg is not an integer, drop it, otherwise return it with "even" or "odd" key.
	if num, err := strconv.Atoi(string(msg)); err != nil {
		return fn.MessagesBuilder().Append(fn.MessageToDrop())
	} else if num%2 == 0 {
		return fn.MessagesBuilder().Append(fn.MessageTo("even", msg))
	} else {
		return fn.MessagesBuilder().Append(fn.MessageTo("odd", msg))
	}
}

func runQuery(cmd *cobra.Command, args []string) error {
	return server.New().RegisterMapper(fn.MapFunc(handle)).Start(context.Background())
}
```

sandangel:
Return error instead of log.Fatal, fix #35

### DIFF
--- a/pkg/function/server/server.go
+++ b/pkg/function/server/server.go
@@ -117,6 +117,7 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) error {
 	functionpb.RegisterUserDefinedFunctionServer(grpcServer, s.svc)
 
 	errCh := make(chan error, 1)
+	defer close(errCh)
 	// start the grpc server
 	go func(ch chan<- error) {
 		log.Println("starting the gRPC server with unix domain socket...")
--- a/pkg/function/server/server.go
+++ b/pkg/function/server/server.go
@@ -128,7 +128,7 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
 	grpcServer.GracefulStop()
 }
 
-// StartE starts the gRPC server via unix domain socket at configs.Addr and return errors.
+// StartE starts the gRPC server via unix domain socket at configs.Addr and return error.
 func (s *server) StartE(ctx context.Context, inputOptions ...Option) error {
 	var opts = &options{
 		sockAddr:       functionsdk.Addr,
--- a/pkg/function/server/server.go
+++ b/pkg/function/server/server.go
@@ -81,7 +81,55 @@ func (s *server) RegisterReducer(r functionsdk.ReduceHandler) *server {
 }
 
 // Start starts the gRPC server via unix domain socket at configs.Addr.
-func (s *server) Start(ctx context.Context, inputOptions ...Option) error {
+func (s *server) Start(ctx context.Context, inputOptions ...Option) {
+	var opts = &options{
+		sockAddr:       functionsdk.Addr,
+		maxMessageSize: functionsdk.DefaultMaxMessageSize,
+	}
+
+	for _, inputOption := range inputOptions {
+		inputOption(opts)
+	}
+
+	cleanup := func() {
+		if _, err := os.Stat(opts.sockAddr); err == nil {
+			if err := os.RemoveAll(opts.sockAddr); err != nil {
+				log.Fatal(err)
+			}
+		}
+	}
+	cleanup()
+
+	ctxWithSignal, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	lis, err := net.Listen(functionsdk.Protocol, opts.sockAddr)
+	if err != nil {
+		log.Fatalf("failed to execute net.Listen(%q, %q): %v", functionsdk.Protocol, functionsdk.Addr, err)
+	}
+	grpcServer := grpc.NewServer(
+		grpc.MaxRecvMsgSize(opts.maxMessageSize),
+		grpc.MaxSendMsgSize(opts.maxMessageSize),
+	)
+	functionpb.RegisterUserDefinedFunctionServer(grpcServer, s.svc)
+
+	// start the grpc server
+	go func() {
+		log.Println("starting the gRPC server with unix domain socket...")
+		err = grpcServer.Serve(lis)
+		if err != nil {
+			log.Fatalf("failed to start the gRPC server: %v", err)
+		}
+	}()
+
+	<-ctxWithSignal.Done()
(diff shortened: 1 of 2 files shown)


pkg/function/server/server.go
```diff
@@ -90,39 +90,52 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
...
-		log.Fatalf("failed to execute net.Listen(%q, %q): %v", functionsdk.Protocol, functionsdk.Addr, err)
+		return err
 	}
 	grpcServer := grpc.NewServer(
 		grpc.MaxRecvMsgSize(opts.maxMessageSize),
 		grpc.MaxSendMsgSize(opts.maxMessageSize),
 	)
 	functionpb.RegisterUserDefinedFunctionServer(grpcServer, s.svc)
 
+	errCh := make(chan error, 1)
```
vigith, line 119: can you close the errCh? perhaps, you can do a `defer` on closing the channel?

vigith (commented):
DCO is missing

pkg/function/server/server.go
sandangel: done

sandangel:
> DCO is missing

I signed off. This check is interesting. My first time experience. ^^

pkg/function/server/server.go
```diff
@@ -90,39 +90,53 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
...
-			if err := os.RemoveAll(opts.sockAddr); err != nil {
-				log.Fatal(err)
-			}
+	cleanup := func() error {
+		_, err := os.Stat(opts.sockAddr)
+		if err == nil {
+			err = os.RemoveAll(opts.sockAddr)
 		}
+		return err
+	}
```
KeranYang, line 99: It seems the logic within cleanup changed a bit. Before the change, if `os.Stat(opts.sockAddr)` returns non-nil error, we ignore it. After the change, we return the error. @vigith @whynowy thoughts?

pkg/function/server/server.go
```diff
@@ -90,39 +90,53 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
...
+
+	err := cleanup()
+	if err != nil {
+		return err
 	}
```
KeranYang, line 104:
```suggestion
        if err := cleanup(); err != nil {
		return err
	}
```

pkg/function/server/server.go
```diff
@@ -90,39 +90,53 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
...
-	cleanup()
 
 	ctxWithSignal, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
 	lis, err := net.Listen(functionsdk.Protocol, opts.sockAddr)
 	if err != nil {
-		log.Fatalf("failed to execute net.Listen(%q, %q): %v", functionsdk.Protocol, functionsdk.Addr, err)
+		return err
```
KeranYang, line 111:
```suggestion
		return fmt.Errorf("failed to execute net.Listen(%q, %q): %v", functionsdk.Protocol, functionsdk.Addr, err)
```

pkg/function/server/server.go
```diff
@@ -90,39 +90,53 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
...
+	errCh := make(chan error, 1)
+	defer close(errCh)
 	// start the grpc server
-	go func() {
+	go func(ch chan<- error) {
 		log.Println("starting the gRPC server with unix domain socket...")
 		err = grpcServer.Serve(lis)
 		if err != nil {
-			log.Fatalf("failed to start the gRPC server: %v", err)
+			ch <- err
```
KeranYang, line 128:
```suggestion
			ch <- fmt.Errorf("failed to start the gRPC server: %v", err)
```

sandangel:
I added `StartE` function as discussed with @whynowy in #35

pkg/sink/server/server.go
```diff
@@ -91,3 +92,61 @@ func (s *server) Start(ctx context.Context, inputOptions ...Option) {
 	defer log.Println("Successfully stopped the gRPC server")
 	grpcServer.GracefulStop()
 }
+
+// StartE starts the gRPC server via unix domain socket at configs.Addr and return error.
+func (s *server) StartE(ctx context.Context, inputOptions ...Option) error {
```
whynowy, line 97: I guess we don't need 2 similar functions for now, keeping `Start() error` is good enough.

sandangel: I updated the code

pkg/function/server/server.go
vigith: yes, file not found should be ignored.

chromevoid:
Could you also add how you test the code change and what the test result is in the PR comment?

sandangel:
I don't see any tests for the Start function tho. Should I add one?

whynowy:
> I don't see any tests for the Start function tho. Should I add one?

Appreciate if you could :)

sandangel:
So the test scenarios would be complex. Like checking if the server is running, check if the error returning, check if the server shutdown gracefully.

whynowy (approved):
LGTM

pkg/function/server/server.go
KeranYang: According to Vigith's reply, suggesting the following change:

```suggestion
	cleanup := func() error {
		if _, err := os.Stat(opts.sockAddr); err == nil {
			return os.RemoveAll(opts.sockAddr)
		}
		return nil
	}
```

vigith (approved):

KeranYang (approved):

PR #37, opened 2023-02-14
feat: bidirectional streaming

### LINKED ISSUE #38: numaflow-go changes to support gRPC bidirectional streaming for reduce

yhl25:
fixes #38 
Signed-off-by: Yashash H L <yashash_hl@intuit.com>


pkg/function/client/client.go
```diff
@@ -79,20 +81,51 @@ func (c *client) MapTFn(ctx context.Context, datum *functionpb.Datum) ([]*functi
...
+	datumList := &functionpb.DatumList{}
+
 	stream, err := c.grpcClt.ReduceFn(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to execute c.grpcClt.ReduceFn(): %w", err)
 	}
-	for datum := range datumStreamCh {
-		if err := stream.Send(datum); err != nil {
-			return nil, fmt.Errorf("failed to execute stream.Send(%v): %w", datum, err)
+	g.Go(func() error {
```
whynowy, line 92: There's only 1 goroutine here, what's the benefit of using errgroup?

pkg/function/client/client.go
```diff
@@ -79,20 +81,51 @@ func (c *client) MapTFn(ctx context.Context, datum *functionpb.Datum) ([]*functi
...
+		var sendErr error
+		for datum := range datumStreamCh {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+				if sendErr = stream.Send(datum); sendErr != nil {
+					return sendErr
```
whynowy, line 100: Does it need to `CloseSend()`?

pkg/function/service.go
```diff
@@ -221,28 +218,40 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
...
-				}
+			func(key string, ch chan Datum) {
+				g.Go(func() error {
+					messages := fs.Reducer.HandleDo(ctx, key, ch, md)
+					datumList := buildDatumList(messages)
+
+					// stream.Send() is not thread safe.
+					mu.Lock()
+					defer mu.Unlock()
+					err := stream.Send(datumList)
```
whynowy, line 229: This is assuming the returned datum list is not over the size limit?

pkg/function/client/client.go
yhl25: even if there is one goroutine we will need a wait group, instead of wait group I am using err group because the goroutine might return an error.

pkg/function/client/client.go
yhl25: No we don't need to. 
I looked at their implementation they call finish on the stream if there are any errors
```
	defer func() {
		if err != nil && err != io.EOF {
			// Call finish on the client stream for errors generated by this SendMsg
			// call, as these indicate problems created by this client.  (Transport
			// errors are converted to an io.EOF error in csAttempt.sendMsg; the real
			// error will be returned from RecvMsg eventually in that case, or be
			// retried.)
			cs.finish(err)
		}
	}()
```

pkg/function/service.go
yhl25: yeah, if it's over the size limit it will return an error

pkg/function/client/client.go
```diff
@@ -79,20 +81,51 @@ func (c *client) MapTFn(ctx context.Context, datum *functionpb.Datum) ([]*functi
...
+	})
+
+outputLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+			var resp *functionpb.DatumList
+			resp, err = stream.Recv()
```
ashwinidulams, line 114: Just a note here: I think this stream is going to be busy once the window closes. so that means the http/2 stream will pack enough datum objects to keep the overhead to a minimum. but if this assumption is not going to hold, then this becomes one of the perf bottleneck.

pkg/function/client/client.go
```diff
@@ -79,20 +81,51 @@ func (c *client) MapTFn(ctx context.Context, datum *functionpb.Datum) ([]*functi
...
-	reducedDatumList, err := stream.CloseAndRecv()
+
+	err = g.Wait()
 	if err != nil {
-		return nil, fmt.Errorf("failed to execute stream.CloseAndRecv(): %w", err)
+		return nil, err
 	}
 
-	return reducedDatumList.GetElements(), nil
+	return datumList, err
```
ashwinidulams, line 130: are we planning to extend the stream abstraction to the forward as well (as against accumulating data) ?

pkg/function/service.go
```diff
@@ -221,28 +218,40 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
...
 
 	// wait until all the HandleDo return
-	wg.Wait()
-	return stream.SendAndClose(&functionpb.DatumList{
-		Elements: datumList,
-	})
+	return g.Wait()
+}
+
+func buildDatumList(messages Messages) *functionpb.DatumList {
```
ashwinidulams, line 245: are we still using this?

pkg/function/client/client.go
vigith: could you please note that in the comment?

pkg/function/client/client.go
vigith: please add a TODO. or in fact, create an issue to track this.

vigith (approved):

ashwinidulams (approved):

KeranYang (approved):

PR #39, opened 2023-03-13
refactor: un-export MessageT's attributes

KeranYang:
In MapT, it's important to assign new event times to ensure that default time values are not mistakenly passed to the next vertex. While our MapT SDKs offer methods like MessageTToAll() and MessageTTo() to create MessageT objects with explicit event times, users can also construct MessageT objects from scratch. However, if they forget to assign a new event time while modifying other attributes, default time values may be used unintentionally. 

To reduce this risk, we propose always requiring an explicit event assignment for MapT implementation. This code change achieves the proposal by un-exporting MessageT's attributes.

After this change, user can only use the following constructors to create MessageT objects:

```
functionsdk.MessageTToDrop()
```

```
functionsdk.MessageTTo(eventTime, key, value)
```

```
functionsdk.MessageTToAll(eventTime, value)
```

### Following approaches to create MessageT objects are no longer allowed.

```
msgt := MessageT{EventTime: event_time, Key: key, Value: value}
```
```
msgt := MessageT{}
msgt.EventTime = event_time
...
```

**Also applied similar changes to Message**

### DIFF
--- a/pkg/function/examples/event_time_filter/impl/filter_test.go
+++ b/pkg/function/examples/event_time_filter/impl/filter_test.go
@@ -61,33 +61,21 @@ func Test_FilterEventTime(t *testing.T) {
 			name:  "DatumWithEventTimeBefore2022GetsDropped",
 			input: beforeYear2022Datum{},
 			expectedOutput: functionsdk.MessageTs{
-				functionsdk.MessageT{
-					EventTime: time.Time{},
-					Key:       functionsdk.DROP,
-					Value:     []byte{},
-				},
+				functionsdk.MessageTToDrop(),
 			},
 		},
 		{
 			name:  "DatumWithEventTimeWithin2022GetsKeyAndEventTimeUpdated",
 			input: withinYear2022Datum{},
 			expectedOutput: functionsdk.MessageTs{
-				functionsdk.MessageT{
-					EventTime: janFisrt2022,
-					Key:       "within_year_2022",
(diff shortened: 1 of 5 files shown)


KeranYang:
@whynowy @vigith I am thinking about making similar changes to `Message` but I am not sure if we have any existing Numaflow users that are already using the `Message{}` approaches to instantiate Message object. Could you guys help confirm?

If we have, we might ask users to make corresponding changes to use `MessageTo*` before un-exporting.

vigith (approved):
sounds good to me

whynowy (approved):
LGTM

whynowy:
> @whynowy @vigith I am thinking about making similar changes to `Message` but I am not sure if we have any existing Numaflow users that are already using the `Message{}` approaches to instantiate Message object. Could you guys help confirm?
> 
> If we have, we might ask users to make corresponding changes to use `MessageTo*` before un-exporting.

Go with it. Ppl will not able to compile with the new change if they do, then they are gonna realize it.

yhl25:
how will the user test the output from handles if it's private? we should think from the unit testing perspective as well

KeranYang:
> how will the user test the output from handles if it's private? we should think from the unit testing perspective as well

For unit test, user will need to use our public constructors to prepare expected output. An example: https://github.com/numaproj/numaflow-go/blob/main/pkg/function/examples/event_time_filter/impl/filter_test.go#L51

PR #42, opened 2023-04-04
feat: introduce info server

whynowy:
# Description

copilot:summary

### DIFF
--- a/pkg/info/types.go
+++ b/pkg/info/types.go
@@ -1,0 +1,28 @@
+package info
+
+const (
+	SocketAddress = "/var/run/numaflow/info.sock"
+)
+
+type Protocol string
+
+const (
+	UDS Protocol = "uds"
+	TCP Protocol = "tcp"
+)
+
+type Language string
+
+const (
+	Go     Language = "go"
+	Python Language = "python"
+	Java   Language = "java"
+)
+
+// ServerInfo is the information about the server
+type ServerInfo struct {
+	Protocol  Protocol          `json:"protocol"`
+	Language  Language          `json:"language"`
+	Version   string            `json:"version"`
+	Metaddata map[string]string `json:"metadata"`
+}
--- a/pkg/function/client/client.go
+++ b/pkg/function/client/client.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
 	"github.com/numaproj/numaflow-go/pkg/function"
+	"github.com/numaproj/numaflow-go/pkg/info"
+	infoclient "github.com/numaproj/numaflow-go/pkg/info/client"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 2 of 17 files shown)


ashwinidulams:
do we have explicit ordering for udf and numa containers today?

pkg/info/types.go
```diff
@@ -0,0 +1,28 @@
...
+	UDS Protocol = "uds"
+	TCP Protocol = "tcp"
+)
+
+type Language string
+
+const (
+	Go     Language = "go"
+	Python Language = "python"
+	Java   Language = "java"
```
vigith, line 19: why do we need to track each language on numaflow-go?

vigith:
The info-server is used only in the bootup; will it have any purpose after that? If this is only for a one-time information exchange, should we just pass it over a file or something as an initial handshake mechanism?

whynowy:
> do we have explicit ordering for udf and numa containers today?

There's no starting order for normal containers, so our numa container does a waituntilready during startup.

pkg/info/types.go
whynowy: Our sdk versions are not consistent. We should have a mapping in numaflow code to track all the versions for compatibility check.

vigith:
> > The info-server is used only in the bootup; will it have any purpose after that? If this is only for a one-time information exchange, should we just pass it over a file or something as an initial handshake mechanism?
> 
> It works with file. Thinking shall we get it integrated with the prometheus support together [numaproj/numaflow#324](https://github.com/numaproj/numaflow/issues/324).

I do not think it is wise to integrate https://github.com/numaproj/numaflow/issues/324 since it is very opinionated. Let's keep this as lightweight as possible so that the resource usage will be minimal and only on the bootup.

pkg/info/server_info.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+	"log"
+	"os"
+	"runtime/debug"
+	"strings"
+	"time"
+)
+
+func GetSDKVersion() string {
+	version := "unknown"
```
vigith, line 15: is an empty string a better value than "unknown"?

pkg/info/server_info.go
```diff
@@ -0,0 +1,121 @@
...
+			} else {
+				break existence
+			}
+		}
+	}
+
+	// Also check if the file is ready to read
+	// It takes some time for the server to write the server info file
+	// TODO: use a better way to wait for the file to be ready
+	b, err := os.ReadFile(options.svrInfoFilePath)
```
vigith, line 73: will a file.stat > 0 length be a good check?

pkg/info/server_info.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+	for len(b) == 0 {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			time.Sleep(100 * time.Millisecond)
+			b, err = os.ReadFile(options.svrInfoFilePath)
```
vigith, line 83: how do we handle partial/in-progress writes, eg. numaflow starting reading early before the entire file has been flushed?

pkg/info/server_info.go
```diff
@@ -0,0 +1,121 @@
...
+func Read(opts ...Option) (*ServerInfo, error) {
+	options := defaultOptions()
+	for _, opt := range opts {
+		opt(options)
+	}
+	// It takes some time for the server to write the server info file
+	// TODO: use a better way to wait for the file to be ready
+	retry := 0
+	b, err := os.ReadFile(options.svrInfoFilePath)
+	for len(b) == 0 && err == nil && retry < 10 {
```
vigith, line 105: will we break the loop too early because we read partial data?

pkg/info/server_info.go
whynowy: updated.

pkg/info/server_info.go
whynowy: Changed to write an END mark.

vigith (approved):

PR #43, opened 2023-04-04
feat: Add ID and numDelivered fields to Datum.

xdevxy:
As part of https://github.com/numaproj/numaflow/issues/538, this PR adds ID and numDelivered fields to Datum, so that users are exposed with these metadata information about the message.


pkg/apis/proto/function/v1/udfunction.proto
```diff
@@ -44,6 +44,8 @@ message Datum {
   bytes value = 2;
   EventTime event_time = 3;
   Watermark watermark = 4;
+  string id = 5;
+  uint64 num_delivered = 6;
```
vigith, line 48: should we create a new type and embed it in Datum?

xdevxy: I'm fine with both. However I do see wrapping them in a new type such as `Metadata` makes it more clear.

xdevxy: Updated.

pkg/apis/proto/sink/v1/udsink.proto
```diff
@@ -35,6 +35,7 @@ message Datum {
   EventTime event_time = 3;
   Watermark watermark = 4;
   string id = 5;
+  uint64 num_delivered = 6;
```
vigith, line 38: let's do the same for Sink Datum too?

pkg/function/interface.go
```diff
@@ -10,9 +10,11 @@ import (
 
 // Datum contains methods to get the payload information.
 type Datum interface {
```
vigith, line 12: perhaps we should split this also into DatumMetadata() and return that interface on which the UDF author can call `.Id()` and `.NumDelivered()`?

vigith (commented):
could you please add a godoc for exported functions?

pkg/apis/proto/sink/v1/udsink.proto
```diff
@@ -34,7 +34,19 @@ message Datum {
   bytes value = 2;
   EventTime event_time = 3;
   Watermark watermark = 4;
-  string id = 5;
+
+  reserved 5;
```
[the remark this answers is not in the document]
vigith: we will support backward compatibility from 1.0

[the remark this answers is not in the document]
xdevxy: Just for better best practice, can we keep it as it won't bring any harm. :)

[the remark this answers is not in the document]
xdevxy: I see, then to be more clean, I will remove it.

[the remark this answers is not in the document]
xdevxy: > you can remove this since we do not support backward compatibility

Which one specific you are talking about? Thanks!

KeranYang:
Have we thought about using `map<string, string>` as Metadata type? Such that we won't need to add attributes to the `Metadata` every time we have new use cases other than `id` and `num_delivered`.

I also noticed we are making metadata changes across numaflow and its SDKs. With a generic `map` data type, will it save us some efforts in the future when it comes to adding more metadatas?

cc: @vigith @whynowy

xdevxy:
> Have we thought about using `map<string, string>` as Metadata type? Such that we won't need to add attributes to the `Metadata` every time we have new use cases other than `id` and `num_delivered`.
> 
> I also noticed we are making metadata changes across numaflow and its SDKs. With a generic `map` data type, will it save us some efforts in the future when it comes to adding more metadata?
> 
> cc: @vigith @whynowy

That's a good point, although not sure how to handle generic type via protobuf, as if the metadata type would vary. If all using `string` we may need to do custom parsing which can be error-prone.

KeranYang:
> we may need to do custom parsing which can be error-prone
My understanding is Numaflow platform will take responsibility of ensuring the strings are in correct format, which is under our control.

Numaflow user will need to parse the data, which requires us to define a metadata contract to make it clear which key should be guaranteed in what format. It's extra work to build such contract with users but I feel it could still be a good option to consider, comparing with making code changes to add metadata.

vigith:
> Have we thought about using `map<string, string>` as Metadata type? Such that we won't need to add attributes to the `Metadata` every time we have new use cases other than `id` and `num_delivered`.
> 
> I also noticed we are making metadata changes across numaflow and its SDKs. With a generic `map` data type, will it save us some efforts in the future when it comes to adding more metadata?
> 
> cc: @vigith @whynowy

That is a fair point, but I think we can revisit this in the future if we are adding more fields to the metadata. It shouldn't break the user contract because we expose only the interface to the user.

KeranYang:
> It shouldn't break the user contract because we expose only the interface to the user.

I see. You mean the `ID()` and `NumDelivered()` interfaces. I was thinking that we have one single interface `GetMetadata() <string, string>` and delegate the parsing work to the user. I think with such implementation, we can achieve 0 changes to SDK packages when adding new metadata in the future. Imagining we want to add `ProcessedByPod` metadata, we just update Numaflow platform to add the <`ProcessedByPod`, `{podName}`> k/v pairs to metadata. On users' side, they can immediately start consuming `ProcessedByPod` metadata without us publishing a new release for SDKs.

I think we should consider this approach if it makes sense. If we move forward with exposing interfaces for each metadata, it will be hard to switch to my proposal in the future, as it's backward-incompatible.

xdevxy:
> > It shouldn't break the user contract because we expose only the interface to the user.
> 
> I see. You mean the `ID()` and `NumDelivered()` interfaces. I was thinking that we have one single interface `GetMetadata() <string, string>` and delegate the parsing work to the user. I think with such implementation, we can achieve 0 changes to SDK packages when adding new metadata in the future. Imagining we want to add `ProcessedByPod` metadata, we just update Numaflow platform to add the <`ProcessedByPod`, `{podName}`> k/v pairs to metadata. On users' side, they can immediately start consuming `ProcessedByPod` metadata without us publishing a new release for SDKs.
> 
> I think we should consider this approach if it makes sense. If we move forward with exposing interfaces for each metadata, it will be hard to switch to my proposal in the future, as it's backward-incompatible.

I'm not sure if we want to have users to do the custom parsing, as that requires we document how to parse each metadata and they need to understand and follow that, even though sometimes it could be trivial.

KeranYang:
> as that requires we document how to parse each metadata and they need to understand and follow that

That's fair. I am ok with moving forward with current approach. :)

whynowy:
Keran's point is very fair. However I can not imagine if there's any other metadata needs to be added in the future, let's keep it as is, and revisit it if we really need to do so.

whynowy (approved):

PR #44, opened 2023-04-05
feat: add multiproc tcp client support

kohlisid:


pkg/function/client/client_resolver.go
```diff
@@ -0,0 +1,76 @@
...
+
+const (
+	exampleScheme      = "example"
+	exampleServiceName = "lb.example.grpc.io"
+	connAddr           = "0.0.0.0"
+)
+
+var addrsList []string
+
+type exampleResolverBuilder struct{}
```
vigith, line 20: rename please

pkg/function/configs.go
```diff
@@ -12,3 +12,19 @@ const (
 	WinStartTime          = "x-numaflow-win-start-time"
 	WinEndTime            = "x-numaflow-win-end-time"
 )
+
+var Protocol string
+var Addr string
+var MAP_MULTIPROC_SERV bool
```
vigith, line 18: let's use getters for these? we might not need these vars

pkg/function/client/client_resolver.go
```diff
@@ -0,0 +1,76 @@
...
+func buildConnAddrs(numCpu int) {
+	var conn = make([]string, numCpu)
+	for i := 0; i < numCpu; i++ {
+		conn[i] = connAddr + function.Addr + "," + strconv.Itoa(i+1)
+	}
+	addrsList = conn
+}
+func init() {
+	if function.MAP_MULTIPROC_SERV == true {
+		resolver.Register(&exampleResolverBuilder{})
```
whynowy, line 71: My understanding this registration is the purpose of this file, right? 
Using `init()` looks tricky, can we expose a function, and get it called in `client.New()` to achieve the registration?

pkg/function/client/client_resolver.go
```diff
@@ -0,0 +1,76 @@
...
+		numCpu := runtime.NumCPU()
```
whynowy, line 72: 1. `NumCPU()` returns the number of vCores of the K8s node, not the vCores assigned to the `numa` container;
2. My understanding is the `udf` container will start multi processors based on the CPU it reads (or ENV variable), which is a different from the one we get here. If we use ENV variable, it should be configured in both containers.

pkg/function/configs.go
```diff
@@ -12,3 +12,19 @@ const (
...
+
+func init() {
```
whynowy, line 20: Instead of using `init()` Move all of these to `client.New()`?

pkg/function/client/client_resolver.go
ashwinidulams: i have used automaxprocs else where to get the cpu allocation - https://github.com/uber-go/automaxprocs

pkg/function/client/client_resolver.go
kohlisid: Done

pkg/function/client/client_resolver.go
kohlisid: Moved to client.New()

pkg/function/client/client_resolver.go
kohlisid: Have updated to use min of automaxprocs vs numCPU
Add env overide as well

pkg/function/configs.go
kohlisid: Moved

pkg/function/configs.go
```diff
@@ -12,3 +10,20 @@ const (
 	WinStartTime          = "x-numaflow-win-start-time"
 	WinEndTime            = "x-numaflow-win-end-time"
 )
+
+var Protocol string
```
whynowy, line 14: How about just defining constants in this file like following, and then in `client.New()`, check the env variable `MAP_MULTIPROC`, and build the different client with these constants?  Using a function like `setupConn()` to change the global variables seems to be weird.

```golang
const (
  TCP = "tcp"
  UDS = "unix"
  UDS_ADDR = "/var/run/numaflow/function.sock"
  TCP_ADDR = ":55551"
)
```

For golang server, it should only use `UDS` and `UDS_ADDR` to start.

kohlisid: Updated using this approach

whynowy:
@kohlisid - can you resolve the conflict?

pkg/function/client/client_resolver.go
```diff
@@ -0,0 +1,67 @@
...
+	"strings"
+)
+
+const (
+	custScheme      = "example"
+	custServiceName = "lb.example.grpc.io"
+	connAddr        = "0.0.0.0"
+)
+
+var addrsList []string
```
whynowy, line 17: As we discussed offline, this can be moved into `multiProcResolverBuilder` as a property.

kohlisid: Updated

pkg/function/client/client.go
```diff
@@ -23,19 +28,43 @@ type client struct {
...
-	sockAddr := fmt.Sprintf("%s:%s", function.Protocol, opts.sockAddr)
-	conn, err := grpc.Dial(sockAddr, grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(opts.maxMessageSize), grpc.MaxCallSendMsgSize(opts.maxMessageSize)))
-
+	var conn *grpc.ClientConn
+	var err error
+	var sockAddr string
+	// Make a TCP connection client for multiprocessing grpc server
+	if function.IsMapMultiProcEnabled() == true {
+		log.Println("Multiprocessing TCP Client ", function.TCP, function.TCP_ADDR)
```
whynowy, line 53: Should use `opts.sockAddr` in the log.

pkg/function/client/client.go
```diff
@@ -23,19 +28,43 @@ type client struct {
...
+		sockAddr = fmt.Sprintf("%s%s", connAddr, opts.sockAddr)
+		conn, err = grpc.Dial(
+			fmt.Sprintf("%s:///%s", custScheme, custServiceName),
+			// This sets the initial load balancing policy as Round Robin
+			grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`),
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+			grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(opts.maxMessageSize), grpc.MaxCallSendMsgSize(opts.maxMessageSize)),
+		)
+	} else {
+		log.Println("UDS Client ", function.UDS, function.UDS_ADDR)
```
whynowy, line 63: Same here, should use `opts.xxx`

pkg/function/client/client.go
kohlisid: Done

pkg/function/client/client.go
kohlisid: Done

whynowy (approved):

pkg/function/client/client_resolver.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"github.com/numaproj/numaflow-go/pkg/function"
+	"google.golang.org/grpc/resolver"
+	"strconv"
+	"strings"
+)
+
+const (
+	custScheme      = "example"
+	custServiceName = "lb.example.grpc.io"
```
vigith, line 12:
```suggestion
	custScheme      = "udf"
	custServiceName = "numaflow.numaproj.io"
```

could you please fix?

vigith (approved):

pkg/function/client/client.go
```diff
@@ -133,3 +162,22 @@ outputLoop:
...
+func regMultProcResolver() {
+	resolver.Register(&multiProcResolverBuilder{})
+	maxProcs := runtime.GOMAXPROCS(0)
+	numCpu := runtime.NumCPU()
+	if maxProcs < numCpu {
+		numCpu = maxProcs
+	}
+	val, present := os.LookupEnv("NUM_CPU_MULTIPROC")
+	if present {
+		numCpu, _ = strconv.Atoi(val)
```
vigith, line 177:
```suggestion
		numCpu, err := strconv.Atoi(val)
		if err || numCpu < 1 {
			numCpu = maxProcs
		}
```

vigith (approved):

PR #45, opened 2023-04-06
feat: introducing tags for conditional forwarding

yhl25:


pkg/apis/proto/sink/v1/udsink.proto
```diff
@@ -29,7 +29,7 @@ message Watermark {
 /**
  * Datum represents a datum element.
  */
-message Datum {
+message DatumRequest {
```
whynowy, line 32: Is this to avoid the naming confusion with the `Datum` interface?

pkg/function/examples/even_odd/main.go
```diff
@@ -16,9 +16,9 @@ func handle(_ context.Context, keys []string, d functionsdk.Datum) functionsdk.M
 	if num, err := strconv.Atoi(string(msg)); err != nil {
 		return functionsdk.MessagesBuilder().Append(functionsdk.MessageToDrop())
 	} else if num%2 == 0 {
-		return functionsdk.MessagesBuilder().Append(functionsdk.MessageTo([]string{"even"}, msg))
+		return functionsdk.MessagesBuilder().Append(functionsdk.NewMessage(msg).WithKeys([]string{"even"}))
```
whynowy, line 19: Is this example used for conditional forwarding? Does it need tags as well?

pkg/function/message_t.go
```diff
@@ -10,36 +10,51 @@ type MessageT struct {
 	eventTime time.Time
 	keys      []string
 	value     []byte
+	tags      []string
+}
+
+// NewMessageT creates a Message with value
+func NewMessageT(value []byte) MessageT {
```
KeranYang, line 17: https://github.com/numaproj/numaflow/pull/668#discussion_r1160221737

pkg/function/examples/even_odd/main.go
yhl25: yeah, I will add tags

KeranYang (approved):

vigith (approved):

PR #50, opened 2023-04-20
chore: fix NewMessageT order

yhl25:


KeranYang (approved):
LGTM. Let's update the corresponding images when making changes to the UDFs.

vigith (approved):

chromevoid:
> LGTM. Let's update the corresponding images when making changes to the UDFs.

Also please update the function call in numaflow as well.

PR #51, opened 2023-04-21
feat: use serv-info data for starting client

kohlisid:


vigith (approved):
NITs

pkg/function/client/client.go
```diff
@@ -58,11 +47,23 @@ func New(inputOptions ...Option) (*client, error) {
 		log.Printf("ServerInfo: %v\n", serverInfo)
 	}
 
+	// Populate connection variables for client connection
+	// based on multiprocessing enabled/disabled
+	if function.IsMapMultiProcEnabled(serverInfo) {
+		err := regMultProcResolver(serverInfo)
+		if err != nil {
```
KeranYang, line 54:
```suggestion
		if err := regMultProcResolver(serverInfo); err != nil {
```

pkg/function/configs.go
```diff
@@ -17,9 +19,9 @@ const (
 	Delimiter             = ":"
 )
 
-func IsMapMultiProcEnabled() bool {
-	val, present := os.LookupEnv("MAP_MULTIPROC")
-	if present && val == "true" {
+func IsMapMultiProcEnabled(svrInfo *info.ServerInfo) bool {
+	protocol := svrInfo.Protocol
+	if protocol == TCP {
```
KeranYang, line 24:
```suggestion
	if protocol := svrInfo.Protocol; protocol == TCP {
```

pkg/function/client/client.go
```diff
@@ -179,24 +180,15 @@ outputLoop:
...
-	}
-	val, present := os.LookupEnv("NUM_CPU_MULTIPROC")
-	if present {
-		num, err := strconv.Atoi(val)
-		if err != nil || num < 1 {
-			numCpu = maxProcs
-		} else {
-			numCpu = num
-		}
+func regMultProcResolver(svrInfo *info.ServerInfo) error {
```
KeranYang, line 183: Before this change, if we can't find `CPU_LIMIT` or can't concert it to a number, we set the `numCpu` to `runtime.GOMAXPROCS(0)` and continue starting the client using TCP with no error.

With this change, if so, we throw and fail the client creation.

For servers with serverInfo {protocol: TCP} but no `CPU_LIMIT` metadata, this could be a backward-incompatible change. Is this as expected?

vigith: there are default values exposed via env field

KeranYang: I think they come from env field and get populated to the `metadata` from [here](https://github.com/numaproj/numaflow-python/pull/89/files#diff-a7e5474397ecc3f74a92417dfb2cc787227c949e645b2edd8c82e4d13b586a04R239) in Python Server.

But for Java and Golang, currently the ServerInfo metadata is an empty map.

kohlisid: The info-server should be exposing the CPU_LIMIT metadata whenever we want to start a TCP server because we want to maintain consistency in the number of server and client instances in the multiproc server. 
 I think if we don't get this info that should indicate an issue with the info-server (ergo an issue in our platform) and thus not start the client.
We could log the error and continue with a default value (maybe 1) in this case. 
@vigith @whynowy What do you guys suggest?

whynowy: I think it's good for now since only python has the chance to start multiprocessing server, which means it won't be called for the reset.

whynowy:
Can you also remove `_ "go.uber.org/automaxprocs"` from the client code?

pkg/function/client/client.go
```diff
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"os"
-	"runtime"
 	"strconv"
 
 	_ "go.uber.org/automaxprocs"
```
whynowy, line 10: remove this.

KeranYang (approved):

pkg/function/client/client.go
```diff
@@ -73,8 +71,8 @@ func New(inputOptions ...Option) (*client, error) {
 			grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(opts.maxMessageSize), grpc.MaxCallSendMsgSize(opts.maxMessageSize)),
 		)
 	} else {
-		log.Println("UDS Client ", function.UDS, opts.sockAddr)
-		sockAddr = fmt.Sprintf("%s:%s", function.UDS, opts.sockAddr)
+		log.Println("UDS Client ", function.UDS, opts.udsSockAddr)
```
vigith, line 74:
```suggestion
		log.Println("UDS Client:", function.UDS, opts.udsSockAddr)
```
does it make sense to print `socketAddr` in the log?.. if yes please fix for multprocessing tcp client too

kohlisid: Done

vigith (approved):

KeranYang (approved):

whynowy (approved):

PR #52, opened 2023-04-24
chore: server info file readiness check in the clients

whynowy:


pkg/function/client/client.go
```diff
@@ -27,40 +28,44 @@ type client struct {
...
+		return nil, fmt.Errorf("failed to wait until server info is ready: %w", err)
+	}
+
 	serverInfo, err := info.Read(info.WithServerInfoFilePath(opts.serverInfoFilePath))
 	if err != nil {
-		// TODO: return nil, err
-		log.Println("Failed to execute info.Read(): ", err)
+		return nil, fmt.Errorf("failed to read server info: %w", err)
 	}
 	// TODO: Use serverInfo to check compatibility and start the right gRPC client.
```
vigith, line 53: Is this todo still relevant?

vigith (approved):

pkg/function/client/client.go
whynowy: updated.

PR #54, opened 2023-05-02
feat: introduce `MapStreamFn` to stream processed messages during map operation

xdevxy:
Closes https://github.com/numaproj/numaflow/issues/723

As part of https://github.com/numaproj/numaflow/issues/623. This introduces a new gRPC streaming API to facilitate stream processed messages during map operation.


pkg/function/service.go
```diff
@@ -92,13 +92,15 @@ func (m *metadata) IntervalWindow() IntervalWindow {
...
+// handler and the reduce operation handler.
 type Service struct {
 	functionpb.UnimplementedUserDefinedFunctionServer
 
-	Mapper  MapHandler
-	MapperT MapTHandler
-	Reducer ReduceHandler
+	Mapper       MapHandler
+	MapperStream MapStreamHandler
+	MapperT      MapTHandler
```
KeranYang, line 102: Do we plan to add MapperTStream for source data transformer?

pkg/function/client/client.go
```diff
@@ -113,6 +113,47 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
...
+	go func() {
+	outputLoop:
+		for {
+			select {
+			case <-ctx.Done():
+				resultErrCh <- ctx.Err()
+				return
+			default:
+				var resp *functionpb.DatumResponseList
+				resp, err = stream.Recv()
```
KeranYang, line 137: nitpicking: You can use `resp, err := stream.Recv()` to save line 136. It also starts a new `err` variable to avoid shadowing.

pkg/function/client/client.go
```diff
@@ -113,6 +113,47 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
...
+				if err == io.EOF {
+					close(resultCh)
+					close(resultErrCh)
+					break outputLoop
```
KeranYang, line 141: What's the difference between `break outputLoop` and `return`?

pkg/function/client/client.go
```diff
@@ -113,6 +113,47 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
 	return mappedDatumList.GetElements(), nil
 }
 
+// MapStreamFn applies a function to each datum element and returns a stream.
+func (c *client) MapStreamFn(ctx context.Context, datum *functionpb.DatumRequest) (<-chan functionpb.DatumResponse, <-chan error) {
+	resultCh := make(chan functionpb.DatumResponse)
+	resultErrCh := make(chan error)
```
KeranYang, line 119: Could you rename `resultErrCh` to `errCh`?

pkg/function/client/client.go
```diff
@@ -113,6 +113,47 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
...
+	stream, err := c.grpcClt.MapStreamFn(ctx, datum)
+	if err != nil {
+		resultErrCh <- fmt.Errorf("failed to execute c.grpcClt.MapStreamFn(): %w", err)
+		close(resultCh)
+		close(resultErrCh)
```
KeranYang, line 124: Should we move the `close` statements to a defer block such that we ensure `MapStreamFn` always closes the channels before it returns?

pkg/function/service.go
xdevxy: I don't think so ATM, as currently the motivation for the streaming case is most for flat map.

pkg/function/client/client.go
xdevxy: Yeah, they both should work for this case.

pkg/function/client/client.go
```diff
@@ -113,6 +113,45 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
...
+			default:
+				resp, err := stream.Recv()
+				if err == io.EOF {
+					return
+				}
+				if err != nil {
+					errCh <- err
+					return
+				}
+				for _, d := range resp.GetElements() {
```
yhl25, line 145: do we really need an error channel? if we just return when we get an error.

yhl25: is it like an error per message? or an error per stream?

yhl25: nvm, I checked the numaflow pr, it's one error per stream. So basically we are redelivering the message if we encounter an error and since the error could be because of any one of the responses from flatmap we need a channel. With dedup in place we are good.

pkg/function/interface.go
```diff
@@ -50,6 +51,12 @@ type MapHandler interface {
 	HandleDo(ctx context.Context, keys []string, datum Datum) Messages
 }
 
+// MapStreamHandler is the interface of map stream function implementation.
+type MapStreamHandler interface {
+	// HandleDo is the function to process each coming message.
```
yhl25, line 56:
```suggestion
	// HandleDo is the function to process each coming message and streams the result back using a channel.
```

yhl25 (approved):
Can you please add an example UDF with streaming map?

pkg/apis/proto/function/v1/udfunction.proto
```diff
@@ -11,6 +11,9 @@ service UserDefinedFunction {
   // MapFn applies a function to each datum request element.
   rpc MapFn(DatumRequest) returns (DatumResponseList);
 
+  // MapStreamFn applies a function to each datum element and returns a stream.
+  rpc MapStreamFn(DatumRequest) returns (stream DatumResponseList);
```
whynowy, line 15: Is this right? Should it be `returns (stream DatumResponse)`?

pkg/function/interface.go
```diff
@@ -50,6 +51,12 @@ type MapHandler interface {
...
+	HandleDo(ctx context.Context, keys []string, datum Datum) <-chan Messages
```
whynowy, line 57: Similarly, define the interface function as:

```golang
HandleDo(ctx context.Context, keys []string, datum Datum, ch <-chan Messages) 
```

Anyone who creates the channel is supposed to close the channel.

whynowy: Also, it should be `ch <-chan Message`, not `Messages`.

whynowy (changes requested):
Please address the comments.

pkg/function/service.go
KeranYang: Sorry I forgot to mention that MapT is a flat map transformer. It's Map + Assigning Event Time to messages.

pkg/function/clienttest/clienttest.go
```diff
@@ -48,32 +48,30 @@ func (c *client) MapFn(ctx context.Context, datum *functionpb.DatumRequest) ([]*
...
-		close(resultErrCh)
-		return resultCh, resultErrCh
+		close(errCh)
+		return resultCh, errCh
 	}
 
 	go func() {
-	outputLoop:
+		defer close(resultCh)
+		defer close(errCh)
```
KeranYang, line 62: Can we move the `defer` statements to right below `make(chan)`?

pkg/function/service.go
KeranYang: Makes sense. Thanks.

pkg/apis/proto/function/v1/udfunction.proto
xdevxy: Put DatumResponseList in case users want to return multiple DatumResponse at a time. But that doesn't really matter as we are streaming.

pkg/function/interface.go
xdevxy: Makes sense, with current logic inside `HandleDo` it has to use another go routine to pass the value to the channel, which is not very idiomatic.

xdevxy:
> Can you please add an example UDF with streaming map?

Sure, I'd like do it in a follow up PR. Thanks!

pkg/function/service.go
```diff
@@ -132,6 +134,42 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.DatumRequest) (*func
...
+		value:     d.GetValue(),
+		eventTime: d.GetEventTime().EventTime.AsTime(),
+		watermark: d.GetWatermark().Watermark.AsTime(),
+		metadata: handlerDatumMetadata{
+			id:           d.GetMetadata().GetId(),
+			numDelivered: d.GetMetadata().GetNumDelivered(),
+		},
+	}
+	ctx := stream.Context()
+	messageCh := make(chan Message)
```
whynowy, line 149: Do we expect the udf developer to close the channel in the end of their `handle()` function?

I think no matter whether the udf close the channel or not, we should support both cases (1. Channel is closed. 2. Function returned without closing the channel). Both of these cases can be considered as the udf function is done. And we should make sure the channel is closed if it's not done by the udf.

Thoughts?

whynowy:
@xdevxy - force-push will mess up the review context, avoid using it if possible.

pkg/function/service.go
yhl25: +1, we should support both.

xdevxy:
> @xdevxy - force-push will mess up the review context, avoid using it if possible.

Tried to avoid force-push as much as possible, but sorry the latest change has some state not correctly committed for the first try.

pkg/function/service.go
```diff
@@ -132,6 +134,44 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.DatumRequest) (*func
...
+
+	go func() {
+		fs.MapperStream.HandleDo(ctx, d.GetKeys(), &hd, messageCh)
+		// Close the channel when not closed by the user.
+		if !isClosed(messageCh) {
```
whynowy, line 154: To my understanding, `isClosed` is very tricky in Golang. An alternative I can think of is, using `for/select` to read messages, and check if the goroutine above is done. Think it through to see what is the best way.

pkg/function/service.go
```diff
@@ -293,3 +333,12 @@ func getValueFromMetadata(md grpcmd.MD, k string) (string, error) {
 	}
 	return value, nil
 }
+
+func isClosed(c chan Message) bool {
+	select {
+	case <-c:
```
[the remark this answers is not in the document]
xdevxy: This can only be used when we know there would be any more data send to the channel, which is exactly the case we are facing. Using this article from go 101 as the reference https://go101.org/article/channel-closing.html.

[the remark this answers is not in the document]
whynowy: 👍 good read..

Question - if reading `messageCh` is slow in line 159-170, which means line 154 might be executed earlier, will that be a problem?

pkg/function/service.go
xdevxy: Yeah, with that we will leave the channel open which probably would be fine as the channel would be GCed, and the data referenced by the channel should also be GCed. Although I would prefer not to do so as it is not a good practice, and there is an alternative to properly close the channel.

pkg/function/service.go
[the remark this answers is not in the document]
xdevxy: Good question, tried this in playground and a slow receiver still can get all data https://go.dev/play/p/oBIyPVmbA4p

pkg/function/service.go
whynowy: I agree, the channel can be GCed, we should focus on not losing any messages. Will following code work?

```golang
	done := make(chan bool)
	go func() {
		fs.MapperStream.HandleDo(ctx, d.GetKeys(), &hd, messageCh)
		done <- true
	}
	finished := false
	for {
		select {
			case <-closeCh:
				finished = true
			case message, ok := <- range messageCh:
				if !ok {
					close(done)
					return nil
				}
				// send message
			default:
				if finished {
					close(messageCh)
					close(done)
					return nil
				}
		}
	}
```

pkg/function/service.go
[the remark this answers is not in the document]
whynowy: No, it missed 6 :)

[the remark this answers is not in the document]
whynowy: Because of the slow consumer, the `isClosed()` function used one message.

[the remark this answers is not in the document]
whynowy: hmmmm...actually, you are right.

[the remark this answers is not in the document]
whynowy: This is interesting, a channel can not be read the data written in the same goroutine...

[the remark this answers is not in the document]
xdevxy: Right, exactly.

pkg/function/service.go
xdevxy: the problem of this is 1) we cannot close the channel in the receiver go routine as `isClosed` won't work as we discussed. 2) without checking if the channel is closed, we cannot close the channel safely. So we have to leave it to GC.

whynowy: > the problem of this is 1) we cannot close the channel in the receiver go routine as `isClosed` won't work as we discussed. 2) without checking if the channel is closed, we cannot close the channel safely. So we have to leave it to GC.

With the code above, we do not need to check if the channel is closed.

2 use cases:

1. Channel is closed by the UDF, then it will end in the branch `case message, ok := <- messageCh:` after reading all the messages;
2. Channel is not closed by the UDF, then it will go to `case <-done` first, and it will go to `default`, and it's safe to `close(messageCh)`.

whynowy: Check the sample code https://go.dev/play/p/JOk8nxyDXYw.

xdevxy: Sounds good! As we discussed offline, even though both approach work, I'm updating it per your suggestions to make it easier to follow. Thanks!

pkg/function/service.go
```diff
@@ -132,6 +134,56 @@ func (fs *Service) MapFn(ctx context.Context, d *functionpb.DatumRequest) (*func
...
+				return nil
+			}
+			element := &functionpb.DatumResponse{
+				Keys:  message.keys,
+				Value: message.value,
+				Tags:  message.tags,
+			}
+			err := stream.Send(element)
+			if err != nil {
+				close(messageCh)
```
whynowy, line 173: Shall we leave it there  for GC maybe? We don't know if it's closed or not.

whynowy (approved):

PR #56, opened 2023-05-10
feat: error handling for mapFn

jy4096:
Add gRPC errors in the server SDK for Reduce. No changes for Map because Map doesn't return any customized gRPC errors.
TODO: Map Stream.

Add gRPC error handling in the client code for both Map, MapT, and Reduce including new customized Error type.


whynowy (approved):

whynowy:
@ashwinidulams

KeranYang:
do we plan to apply it to MapT as well?

pkg/function/service.go
```diff
@@ -293,10 +300,11 @@ func (fs *Service) ReduceFn(stream functionpb.UserDefinedFunction_ReduceFnServer
...
 					defer mu.Unlock()
-					err := stream.Send(datumList)
-					if err != nil {
-						// TODO: research on gRPC errors and revisit the error handler
-						return err
+					sendErr := stream.Send(datumList)
+					if sendErr != nil {
+						// the error here is returned by stream.Send()
+						// it's already a gRPC error
+						return recvErr
```
whynowy, line 307: return `recvErr`?

pkg/function/client/client.go
```diff
@@ -200,17 +202,44 @@ outputLoop:
...
+		log.Printf("failed %s: %s", name, udfError.Error())
+		return fmt.Errorf("failed %s: %w", name, udfError)
+	}
+	switch statusCode.Code() {
+	case codes.OK:
+		return nil
+	case codes.DeadlineExceeded, codes.Unavailable, codes.Unknown:
+		udfError.errKind = Retryable
+		log.Printf("failed %s: %s", name, udfError.Error())
+		return fmt.Errorf("failed %s: %w", name, udfError)
```
vigith, line 233: why not return `udfError`?

pkg/function/clienttest/error_test.go
vigith, line 1: reuse

xdevxy (commented):
While looking at the Java server side sdk error handling, I noticed we should panic if there is any application error resulting in restart of numa container per the discussion in https://github.com/numaproj/numaflow/issues/232#issuecomment-1513286370?  @jy4096

jy4096:
As the initial error handling implementation we decided to restart for all types of errors from reduce in the numaflow main container as for now.
Check this TODO: https://github.com/numaproj/numaflow/pull/744/files#diff-0c876de13c4addba8242fe3b40e5fd1eb0f3daab316cf0cbb51eb6c3cd1b8ba0R299
We still want to keep retryable type of error in case we may want to handle reduce retryable error in the future if we have to.

PR #57, opened 2023-05-11
chore: add error test case for map streaming UDF client

xdevxy:


pkg/function/clienttest/client_test.go
```diff
@@ -148,6 +150,55 @@ func TestMapStreamFn(t *testing.T) {
...
+	datumCh := make(chan *functionpb.DatumResponse)
+	errs, ctx := errgroup.WithContext(ctx)
+	errs.Go(func() error {
+		return testClient.MapStreamFn(ctx, requestDatum, datumCh)
+	})
+
+	datumResponses := make([]*functionpb.DatumResponse, 0)
+	for msg := range datumCh {
+		datumResponses = append(datumResponses, msg)
+	}
```
KeranYang, line 196: Do we use `datumResponses` in this unit test? Should we verify the content of it?

xdevxy: For this particular test, we don't need to verify the content, the happy case we do. It is needed only for receiving all messages that are already sent to the channel.

KeranYang (approved):

PR #58, opened 2023-05-15
chore: add uuid to metadata

yhl25:


xdevxy (commented):
Wondering why do we need to add the uuid to metadata?

vigith:
we will not be pursuing this.

PR #61, opened 2023-06-06
chore: move client from SDK to numaflow

kohlisid:


pkg/sink/interface.go
```diff
@@ -16,13 +13,6 @@ type Datum interface {
 	ID() string
 }
 
-// Client contains methods to call a gRPC client.
-type Client interface {
```
jy4096, line 20: do we have the Client interface for function? I don't see in the change files...

pkg/sink/server/server_test.go
```diff
@@ -30,74 +26,48 @@ func Test_server_sink(t *testing.T) {
...
+		return result
+	})
+	// note: using actual UDS connection
+	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Second)
+	go func() {
+		time.Sleep(3 * time.Second)
+		cancel()
+	}()
+	err = New().RegisterSinker(sinkHandler).Start(ctx, WithSockAddr(socketFile.Name()), WithServerInfoFilePath(serverInfoFile.Name()))
+	//err := New().RegisterMapper(mapHandler).Start(ctx, WithSockAddr(socketFile.Name()), WithServerInfoFilePath(serverInfoFile.Name()))
```
jy4096, line 49: clean up

pkg/sink/interface.go
kohlisid: Updated

pkg/sink/server/server_test.go
kohlisid: Done

jy4096 (approved):
Looks good to me

yhl25 (approved):

PR #64, opened 2023-08-02
Example UDF for a Map Vertex to test out cycles

juliev0:
This example can be used with an e2e test to verify that cycles are working. 

It imitates failing and needing to retry a given message the first two times it sees it and then succeeding the third time (assuming every message is uniquely identifiable by its content).

### DIFF
--- a/pkg/function/examples/retry/main.go
+++ b/pkg/function/examples/retry/main.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	functionsdk "github.com/numaproj/numaflow-go/pkg/function"
+	"github.com/numaproj/numaflow-go/pkg/function/server"
+)
+
+var retryCounts map[string]int
+
+const SUCCESS_ITERATION = 3
+
+// This imitates a function which fails the first 2 times it sees a message, but then
+// succeeds on the third time.
+// When it fails, it applies a "retry" Tag to the outgoing message, which causes the message to cycle back to it
+// (this is designed to be used as a Map Vertex which Conditionally Forwards back to
+// itself when the "retry" Tag is set, and forwards onto the next Vertex when it's not set).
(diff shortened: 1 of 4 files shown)


pkg/function/examples/retry/main.go
```diff
@@ -0,0 +1,46 @@
...
+// When it fails, it applies a "retry" Tag to the outgoing message, which causes the message to cycle back to it
+// (this is designed to be used as a Map Vertex which Conditionally Forwards back to
+// itself when the "retry" Tag is set, and forwards onto the next Vertex when it's not set).
+func handle(_ context.Context, keys []string, d functionsdk.Datum) functionsdk.Messages {
+	msgBytes := d.Value()
+	msg := string(msgBytes)
+
+	_, found := counts[msg]
+	if !found {
+		counts[msg] = 1
```
KeranYang, line 26: This is a bit nit-picking: 1. In a PROD context, we might have multiple messages sharing the same payload value, making the counts incorrect and 2. go map is not thread safe.

But it should be ok considering we are going to use it for E2E tests, which is light weight.

KeranYang (approved):

pkg/function/examples/retry/main.go
juliev0: 1. Yeah, this doesn’t represent a realistic example that anyone would actually use
2. Oh, is this called by multiple goroutines? I didn’t realize
That also reminds me that this would only make sense if there’s a single replica for the Vertex. I will consider if there’s a better example I could have for the e2e test

KeranYang: > Is this called by multiple goroutines?

Yes, see https://github.com/numaproj/numaflow/blob/main/pkg/forward/forward.go#L287

vigith: please use a lock on the map else it will bite the users who blindly copy-paste the code.

juliev0: now using sync.Map

PR #65, opened 2023-08-03
feat: add proto file for user defined source

KeranYang:
This PR introduces the initial .proto file for the user-defined source. The auto-generated files are generated locally by running the latest `protoc` with version `libprotoc 23.4`


pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,139 @@
...
+
+  // Ack acknowledges a list of datum offsets.
+  // It indicates that the datum stream has been processed by the source vertex.
+  rpc Ack(AckRequest) returns (AckResponse);
+
+  // Pending returns the number of pending records at the user defined source.
+  rpc Pending(google.protobuf.Empty) returns (PendingResponse);
+
+  // IsReady is the heartbeat endpoint for gRPC.
+  rpc IsReady(google.protobuf.Empty) returns (ReadyResponse);
```
vigith, line 23: Let's name it IsSourceReady because eventually, we will also have Source Transformer, and it might need a different health check endpoint since both of them can be used independently or together.

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,139 @@
...
+    // status is the status of the AckResponse.
+    AckStatus status = 1;
+    // non_acked_offsets is the list of offsets that are not acknowledged.
+    repeated Offset non_acked_offsets = 2;
+  }
+  Result result = 1;
+  /**
+   * AckStatus is the status of the AckResponse.
+   */
+  enum AckStatus {
```
vigith, line 102: what will be the behavior for different AckStatus values? Should we make it either all successful or none? How will the source contract look like in the case of partically_acked? Could you please document those too?

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,139 @@
...
+    repeated Offset offsets = 1;
+  }
+  Request request = 1;
+}
+
+/**
+ * AckResponse is the response for acknowledging datum.
+ */
+message AckResponse {
+  message Result {
```
vigith, line 92: won't it be easier if it is a `repeated {Status, Offset}` so it is easy to figure out what is ack'ed and not?

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,139 @@
...
+    EventTime event_time = 4;
+  }
+  Result result = 1;
+}
+
+/**
+ * AckRequest is the request for acknowledging datum.
+ * It takes a list of offsets to be acknowledged.
+ */
+message AckRequest {
```
vigith, line 81: till now, we prefix with `xxxList` if it is a repeated item, i am not opinionated on either, but we should agree and follow the same conventions.

KeranYang: I have strong opinions on not using `xxxList` because it leaves users the impression that the message only contains one single `repeated` field. In the future, we might want to add more fields to enrich the message. e.g. (might not completely making sense) for `AckRequest`, if we want to ack a list of offsets that belong to one of the partitions from the source, we might want to add to the Request message another field `string partition_id = 2`. For that case, I think `Request` is a better name than `RequestList`.

pkg/apis/proto/source/v1/udsource.proto
KeranYang: Thanks for raising this. I need some help with defining the Ack response. This reply is for both this comment the one below.

Our current implementation of source Ack is that the Ack interface can return partial success and data forwarder will keep retrying the failed offsets until the entire batch is Acked. Hence to mimic the same behaviour, I am allowing the udsource to partially ack. If some of the offsets aren't acked, the source forwarder keep initiating new Ack() gRPC call to the udsource until the batch are all successfully acked.

With that, I also feel the AckStatus is redundant. As the caller can derive such information by simply looking at the size of `non_acked_offsets`. Do you think I should completely remove the AckStatus?

KeranYang: I am removing the AckStatus for now as it can be derived from `non_acked_offsets`.

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,127 @@
...
+    // elements is a list of datum response elements.
+    repeated DatumResponse elements = 1;
+  }
+  Result result = 1;
+}
+
+/*
+ * DatumResponse represents a datum response element.
+ */
+message DatumResponse {
```
jy4096, line 58: I see that we have many [nested types](https://protobuf.dev/programming-guides/proto3/#nested) in this design.. looks like it's used if we have an inner type needs to be `repeated` or just to group the types together... (in my understanding..)

in that case... would we consider moving this type inside line 48?

KeranYang: @jy4096 I don't have a good reference for when to use nested types in general. The way I apply nested type is that if I am sure type A is going to be ONLY used by type B, I put type A as an inner type of type B.

Hence you can see I apply the nested type approach mostly for the `Response` and `Request` messages. Regarding `DatumResponse`, currently, it's only used by `ReadResponse`, however I didn't nest it because I want to leave the door open for future use cases. e.g. we might want to add a new API ReadFromPartition(int partitionID) which reads from a specified ID and return a list of `DatumResponse`.

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,127 @@
...
+    bytes payload = 1;
+    // offset is the offset information of the datum.
+    Offset offset = 2;
+    // keys is an optional list of keys associated with the datum.
+    // Key is the "key" attribute in (key,value) as in the map-reduce paradigm.
+    // We add this optional field to support the use case where the user defined source can provide keys for the datum.
+    // e.g. Kafka and Redis Stream message usually include information about the keys.
+    repeated string keys = 3;
+    // event_time is the time associated with each datum.
+    // We add this optional field to support the use case where the user defined source can provide event time for the datum.
```
jy4096, line 70: I think event time is required in order to calculate watermark...

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,138 @@
...
+  google.protobuf.Timestamp event_time = 1;
+}
+
+/*
+ * ReadRequest is the request for reading datum stream from user defined source.
+ */
+message ReadRequest {
+  message Request {
+    // Required field indicating the number of records to read.
+    uint64 num_records = 1;
```
whynowy, line 39: Shall we accommodate the options such as `readTimeOut`?

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,138 @@
...
+ */
+message EventTime {
+  google.protobuf.Timestamp event_time = 1;
+}
+
+/*
+ * ReadRequest is the request for reading datum stream from user defined source.
+ */
+message ReadRequest {
+  message Request {
```
whynowy, line 37: Do we need to define a struct `ReadRequest` which only contains a single field `Request`?  Why not using a one layer struct? This is a common question for all the similar definitions in this PR.

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,138 @@
+syntax = "proto3";
```
whynowy, line 1: Is my understanding correct, all these definitions are only for underlining gRPC calls, we will not use them as the interfaces required for the users.

pkg/apis/proto/source/v1/udsource.proto
vigith: I am thinking about what will be the most efficient way because Read and Ack are in the hot code path. Deserializing offset every time is heavy on CPU. What if we return an array of the same len as items returned by the Read and each index is a bool that says whether the Ack was successful or not?

KeranYang: @vigith do we de-serialize on the platform side during ack? My understanding is that the platform sends N offsets to UDSource, UDSource deserializes them and acks them partially. After that, UDSource returns the failed offsets to the platform and the platform re-sends them back to retry without deserializing. Hence, the platform doesn't perform deserialization during ack. UDSource can't avoid deserialization to ack all the offsets eventually.

If my understanding is correct, how does returning a boolean array help? 🤔

pkg/apis/proto/source/v1/udsource.proto
KeranYang: That's a good point. I am thinking we can start from simple and add more options at a later time. WDYT?

pkg/apis/proto/source/v1/udsource.proto
KeranYang: Thanks! I followed the proto3 API guideline https://protobuf.dev/programming-guides/api/#dont-include-primitive-types

pkg/apis/proto/source/v1/udsource.proto
KeranYang: Yes.

pkg/apis/proto/source/v1/udsource.proto
KeranYang: The boolean array requires the server side to maintain the order, which could be extra effort.

vigith: > UDSource can't avoid deserialization to ack all the offsets eventually.

What if UDSource maintains the offsets in its original form?

pkg/apis/proto/source/v1/udsource.proto
whynowy: Good to know!

pkg/apis/proto/source/v1/udsource.proto
whynowy: The timeout is something required for reading logic, even it's marked as optional in our code, just because there's a default value. Without it, anyone who implements a udsource will have to give a timeout, with this, I think we should put it in the protocol, and we will make sure it always has a value.

pkg/apis/proto/source/v1/udsource.proto
KeranYang: @vigith Thanks for bringing up the idea of using indices. I spent some time diving deeper into the problem and compared the two approaches and wrote the following notes. Could you help me review it and see if it makes sense?

<!DOCTYPE html><h3 cid="n47" mdtype="heading" class="md-end-block md-heading md-focus" style="box-sizing: border-box; white-space: pre-wrap; break-after: avoid-page; break-inside: avoid; orphans: 4; font-size: 1.5em; margin-top: 1rem; margin-bottom: 1rem; position: relative; font-weight: bold; line-height: 1.43; cursor: text; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-style: normal; font-variant-caps: normal; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain md-expand" style="box-sizing: border-box;">Facts</span></h3><ol class="ol-list" cid="n15" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n17" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n18" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">ACK is a blocking call with a timeout.</span></p></li></ol><h3 cid="n19" mdtype="heading" class="md-end-block md-heading" style="box-sizing: border-box; white-space: pre-wrap; break-after: avoid-page; break-inside: avoid; orphans: 4; font-size: 1.5em; margin-top: 1rem; margin-bottom: 1rem; position: relative; font-weight: bold; line-height: 1.43; cursor: text; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-style: normal; font-variant-caps: normal; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Couple of topics to be discussed before merging this PR:</span></h3><ol class="ol-list" cid="n2" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n4" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n5" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Do we support partially successful ACK?</span></strong></span></p></li></ol><p cid="n6" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">I think we should, otherwise, the server side will need to maintain the status of either ALL SUCCESS or ALL FAILURE. This could be extra work for UDSource. One of the cases is that when one of the offsets failed the ack, UDSource has to NoAck the successful ones before returning, which is introducing more calls to the source. An even worse case could be that the request times out and immediately returns the ALL FAILURE result, leaving no time for UDSource to NoAck processed offsets.</span></p><ol class="ol-list" start="2" cid="n23" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n24" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n25" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">What will be the behaviour on the platform side to handle partial success?</span></strong></span></p></li></ol><p cid="n26" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Retry until all success.</span></p><ol class="ol-list" start="3" cid="n30" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n32" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n33" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">What does ACK return?</span></strong></span></p></li></ol><p cid="n34" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">A list of failed offsets. Maybe each with a failure reason code.</span></p><ol class="ol-list" start="4" cid="n38" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n40" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n41" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s " style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">How do we represent the failed offsets?</span></strong></span></p></li></ol><p cid="n80" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Indices v.s. Serialized Offsets (Proposing using Serialized Offsets)</span></p><figure class="md-table-fig md-focus" cid="n57" mdtype="table" style="box-sizing: border-box; margin: 1.2em 0px; overflow-x: auto; max-width: calc(100% + 16px); padding: 0px; cursor: default; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;">

Scenarios | Using the indices from the ReadMessages array | Using the serialized offset byte arrays from the ReadMessages array | Comments
-- | -- | -- | --
A successful batch ACK with the platform retry | UDSource needs to maintain the offsets in their original form in memory. | UDSource needs to serialize/deserialize the offsets, which can be heavy on the CPU. | If using indices, we don't need the concept of offsets in our protobuf definitions. The READ API can simply return messages associated with indices and ACK API acknowledges the indices.
The platform restarts with partial success ACK. (pod restart) | The platform can't retry the failed offsets because it loses track of indices. UDSource refuses to read a new batch because the previous batch hasn't been completely ACKed. | The platform can't retry the failed offsets because it loses track of the offsets. UDSource starts reading a new batch, which can include the previously failed messages because their offsets were not ACKed. | It seems using serialized offsets is more fault-tolerant than using indices.

</figure>

KeranYang: @vigith Thanks for bringing up the idea of using indices for ACKing the messages. After investigating the problem a little bit deeper, I would like to propose continuing to use the serialized offsets approach. Please help me review the following notes.

<!DOCTYPE html><h3 cid="n47" mdtype="heading" class="md-end-block md-heading md-focus" style="box-sizing: border-box; white-space: pre-wrap; break-after: avoid-page; break-inside: avoid; orphans: 4; font-size: 1.5em; margin-top: 1rem; margin-bottom: 1rem; position: relative; font-weight: bold; line-height: 1.43; cursor: text; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-style: normal; font-variant-caps: normal; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain md-expand" style="box-sizing: border-box;">Facts</span></h3><ol class="ol-list" cid="n15" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n17" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n18" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">ACK is a blocking call with a timeout.</span></p></li></ol><h3 cid="n19" mdtype="heading" class="md-end-block md-heading" style="box-sizing: border-box; white-space: pre-wrap; break-after: avoid-page; break-inside: avoid; orphans: 4; font-size: 1.5em; margin-top: 1rem; margin-bottom: 1rem; position: relative; font-weight: bold; line-height: 1.43; cursor: text; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-style: normal; font-variant-caps: normal; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Couple of topics to be discussed before merging this PR:</span></h3><ol class="ol-list" cid="n2" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n4" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n5" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Do we support partially successful ACK?</span></strong></span></p></li></ol><p cid="n6" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">I think we should, otherwise, the server side will need to maintain the status of either ALL SUCCESS or ALL FAILURE. This could be extra work for UDSource. One of the cases is that when one of the offsets failed the ack, UDSource has to NoAck the successful ones before returning, which is introducing more calls to the source. An even worse case could be that the request times out and immediately returns the ALL FAILURE result, leaving no time for UDSource to NoAck processed offsets.</span></p><ol class="ol-list" start="2" cid="n23" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n24" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n25" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">What will be the behaviour on the platform side to handle partial success?</span></strong></span></p></li></ol><p cid="n26" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Retry until all success.</span></p><ol class="ol-list" start="3" cid="n30" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n32" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n33" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">What does ACK return?</span></strong></span></p></li></ol><p cid="n34" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">A list of failed offsets. Maybe each with a failure reason code.</span></p><ol class="ol-list" start="4" cid="n38" mdtype="list" style="box-sizing: border-box; margin: 0.8em 0px; padding-left: 30px; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><li class="md-list-item" cid="n40" mdtype="list_item" style="box-sizing: border-box; margin: 0px; position: relative;"><p cid="n41" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0px 0px 0.5rem; white-space: pre-wrap; position: relative;"><span md-inline="strong" class="md-pair-s " style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">How do we represent the failed offsets?</span></strong></span></p></li></ol><p cid="n80" mdtype="paragraph" class="md-end-block md-p" style="box-sizing: border-box; line-height: inherit; orphans: 4; margin: 0.8em 0px; white-space: pre-wrap; position: relative; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Indices v.s. Serialized Offsets (</span><span md-inline="strong" class="md-pair-s" style="box-sizing: border-box;"><strong style="box-sizing: border-box;"><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">Proposing Serialized Offsets</span></strong></span><span md-inline="plain" class="md-plain" style="box-sizing: border-box;">)</span></p><figure class="md-table-fig md-focus" cid="n57" mdtype="table" style="box-sizing: border-box; margin: 1.2em 0px; overflow-x: auto; max-width: calc(100% + 16px); padding: 0px; cursor: default; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;">

Scenarios | Using the indices from the ReadMessages array | Using the serialized offset byte arrays from the ReadMessages array | Comments
-- | -- | -- | --
A successful batch ACK with the platform retry. | UDSource needs to maintain the offsets in their original form in memory. UDSource becomes stateful. | UDSource needs to serialize/deserialize the offsets, which can be heavy on the CPU. (How heavy?) | If using indices, we don't need the concept of offsets in our protobuf definitions. The READ API can simply return messages associated with indices and ACK API acknowledges the indices.
The platform restarts with partial success ACK. (pod restart) | The platform can't retry the failed offsets because it loses track of indices. UDSource starts reading a new batch, which can include the previously failed messages because their offsets were not ACKed. | The platform can't retry the failed offsets because it loses track of the offsets. UDSource starts reading a new batch, which can include the previously failed messages because their offsets were not ACKed. | Both approach is fault-tolerant.
(Hypothetically) The platform by accident loses track of one of the offsets before ACK. | The platform sends N-1 indices to UDSource. UDSource acks all of them. The platform starts READing a new batch. UDSource refuses to read because it still has an unacked offset in memory. | The platform sends N-1 offsets to UDSource. UDSource acks all of them. The platform starts READing a new batch. UDSource picks up the one offset that was lost from the previous batch and sends it back to the platform to reprocess. | The approach of using indices requires the platform to ensure that for each batch, ALL the read offsets get sent to ACK. While the serialized offset approach doesn't and supports at-least-once semantics.

</figure>

KeranYang: Had an offline discussion with Vigith. We agreed on most of the notes above. The main change is that in the case of ACK failed/timeout, instead of starting a new forwardAChunk call, we agree that the vertex should stop the forwarder. With that being agreed on, the third scenario supports the indices approach.

The decision is that we are going to provide the users with both serialized offsets and indices and let the users decide which one to use, to enforce the must-ack-all contract.

pkg/apis/proto/source/v1/udsource.proto
vigith: now that each will be a separate service, we can call it `IsReady` itself.

vigith (dismissed):

pkg/apis/proto/source/v1/udsource.proto
whynowy: This requires the ack response to give the index of origin position?

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -88,19 +97,36 @@ message AckRequest {
...
+  }
+  // FailedAcknowledgement is the acknowledgement result for a single offset.
+  // It is used by the platform to retry the acknowledgement.
+  message FailedAcknowledgement {
+    // Required field holding the index of the offset in the original batch.
+    uint64 index_at_original_batch = 1;
+    // Required field holding the offset that this ack result corresponds to.
+    Offset offset = 2;
+    // Required field holding the status of the acknowledgement.
+    Status status = 3;
```
jy4096, line 113: (tiny comment): if the status is only used in `FailedAcknowledgement`, then the status must be some kind of error.... I'm thinking if the name `StatusError` is too general?....

pkg/apis/proto/source/v1/udsource.proto
KeranYang: Yes, exactly.

pkg/apis/proto/source/v1/udsource.proto
KeranYang: Right, there can be multiple reasons causing the error. I am currently using a single `StatusError` but leaving the door open for more detailed types. e.g. `StatusTimeoutError` indicating the offset was not acked because of request timeout, `StatusAckError` indicating the offset was not acked because the source API failed to do so.

Let me add them.

jy4096: I think the error message field below is good enough?... may be we don't really need the status

vigith: let's do status as bool, successful, or failed. let's not try to add different statuses because that is an unbounded list.

KeranYang: Since the Ack API only returns failed acknowledgements, I don't see the status being successful. I think we can remove the status for now.

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,142 @@
...
+
+service UserDefinedSource {
+  // Read returns a stream of datum responses.
+  // The size of the returned ReadResponse is less than or equal to the num_records specified in ReadRequest.
+  // If the request timeout is reached on server side, the returned ReadResponse will contain all the datum responses that have been read.
+  rpc Read(ReadRequest) returns (stream ReadResponse);
+
+  // Ack acknowledges a list of datum offsets.
+  // It indicates that the datum stream has been processed by the source vertex.
+  rpc Ack(AckRequest) returns (AckResponse);
```
vigith, line 18:
```suggestion
  // Ack acknowledges a list of datum offsets.
  // When Ack is called, it implicitly indicates that the datum stream has been processed by the source vertex.
  // The caller (numa) expects the Ack to be successful, and it does not expect any errors. 
  // If there are some irrecoverable errors when the callee (UDSource) is processing the Ack,
  // then it is best to crash because there are no other retry mechanisms possible. 
  rpc Ack(AckRequest) returns (AckResponse);
```

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,142 @@
...
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/empty.proto";
+
+package source.v1;
+
+service UserDefinedSource {
+  // Read returns a stream of datum responses.
+  // The size of the returned ReadResponse is less than or equal to the num_records specified in ReadRequest.
+  // If the request timeout is reached on server side, the returned ReadResponse will contain all the datum responses that have been read.
```
vigith, line 13:
```suggestion
  // If the request timeout is reached on server side, the returned ReadResponse will contain all the datum responses that have been read (which could be an empty list).
```

pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -0,0 +1,142 @@
...
+ */
+message AckRequest {
+  message Request {
+    // Required field holding a list of offsets to be acknowledged.
+    // The offsets must be strictly corresponding to the previously read batch,
+    // meaning the offsets must be in the same order as the datum responses in the ReadResponse.
+    // By enforcing ordering, we can save deserialization effort on the server side, assuming the server keeps a local copy of the raw/un-serialized offsets.
+    repeated Offset offset = 1;
+  }
+  // Required field holding the request.
```
vigith, line 86:
```suggestion
  // Required field holding the request. The list will be ordered and will have the same order as the original Read response.
```

vigith (approved):
few NITs

pkg/apis/proto/source/v1/udsource.proto
xdevxy: Discussed this offline, when it is an empty list, should we consider adding another configuration in numaflow to see if continuously reading empty data should raise an alert?

PR #66, opened 2023-08-04
feat: Add Side input proto specifications

kohlisid:
Note: we don't have a key field here because each side input manager is designed to manage only one side input; therefore, the "key" (side input name defined in the pipeline yaml file) is already bind to the side input manager.


pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message EventTime {
+  // event_time is the time associated with each datum.
+  google.protobuf.Timestamp event_time = 1;
+  // future we can add an enum to represent whether the event time is
+  // based on INGEST_TIME or EVENT_TIME or PROCESSING_TIME
+}
+/**
+ * SideInputResponse represents a datum response element.
```
jy4096, line 25: `a datum response element` ??

pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,37 @@
...
+  rpc IsReady(google.protobuf.Empty) returns (ReadyResponse);
+}
+
+message EventTime {
+  // event_time is the time associated with each datum.
+  google.protobuf.Timestamp event_time = 1;
+  // future we can add an enum to represent whether the event time is
+  // based on INGEST_TIME or EVENT_TIME or PROCESSING_TIME
+}
+/**
```
jy4096, line 24: nit: newline between line 23 and 24

jy4096 (changes requested):

pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,37 @@
...
+service UserDefinedSideInput {
+  // SideInputFn applies a function to update the side-input values.
+  rpc RetrieveSideInput(google.protobuf.Empty) returns (SideInputResponse);
+
+  // IsReady is the heartbeat endpoint for gRPC.
+  rpc IsReady(google.protobuf.Empty) returns (ReadyResponse);
+}
+
+message EventTime {
+  // event_time is the time associated with each datum.
```
jy4096, line 19: `with each datum`

pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,37 @@
...
+
+option go_package = "github.com/numaproj/numaflow-go/pkg/apis/proto/sideinput/v1";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/empty.proto";
+
+package sideinput.v1;
+
+service UserDefinedSideInput {
+  // SideInputFn applies a function to update the side-input values.
```
jy4096, line 11:
```suggestion
  // RetrieveSideInput applies a function to retrieve the latest the side-input values.
```

jy4096 (approved):

pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+option go_package = "github.com/numaproj/numaflow-go/pkg/apis/proto/sideinput/v1";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/empty.proto";
+
+package sideinput.v1;
+
+service UserDefinedSideInput {
```
vigith, line 10:
```suggestion
UserDefinedSideInput is used to broadcast changes to all the vertices ...... <fill it up>..
service UserDefinedSideInput {
```

Could you please document the expected behavior and use of this proto file? You may link to other documents.

vigith (changes requested):

pkg/apis/proto/sideinput/v1/sideinput.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * SideInputResponse represents a response to a given side input retrieval request.
+ */
+message SideInputResponse {
+  // value is the side-input payload
+  bytes value = 1;
+  // event_time represents the timestamp at which value is updated
+  EventTime event_time = 2;
```
whynowy, line 37: Do we need this?

The call to `RetrieveSideInput` is supposed to be called by the manager sidecar, I would assume for most of the time, there's no way for the developer to determine the time it got updated, and current time would be used for it.

I guess it is introduced for the purpose of determine if there's any change for the data? I think we should use different approaches for it if it's for that.

vigith: +1, we should remove this as it might cause confusion.

vigith (approved):

jy4096 (approved):

whynowy (approved):

PR #68, opened 2023-08-08
chore: apply small cleanups

KeranYang:
This PR applies some small cleanups that I discovered while I explore the code base.

### DIFF
--- a/pkg/function/configs.go
+++ b/pkg/function/configs.go
@@ -1,10 +1,10 @@
 package function
 
 const (
-	TCP      = "tcp"
-	UDS      = "unix"
-	UDS_ADDR = "/var/run/numaflow/function.sock"
-	TCP_ADDR = ":55551"
+	TCP     = "tcp"
+	UDS     = "unix"
+	UdsAddr = "/var/run/numaflow/function.sock"
+	TcpAddr = ":55551"
 	// DefaultMaxMessageSize overrides gRPC max message size configuration
 	// https://github.com/grpc/grpc-go/blob/master/server.go#L58-L59
 	//   - defaultServerMaxReceiveMessageSize
--- a/pkg/function/server/server.go
+++ b/pkg/function/server/server.go
@@ -9,10 +9,11 @@ import (
 	"os/signal"
 	"syscall"
 
+	"google.golang.org/grpc"
+
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
 	functionsdk "github.com/numaproj/numaflow-go/pkg/function"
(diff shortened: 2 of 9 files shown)


pkg/function/configs.go
```diff
@@ -1,10 +1,10 @@
 package function
 
 const (
-	TCP      = "tcp"
-	UDS      = "unix"
-	UDS_ADDR = "/var/run/numaflow/function.sock"
-	TCP_ADDR = ":55551"
+	TCP     = "tcp"
+	UDS     = "unix"
+	UdsAddr = "/var/run/numaflow/function.sock"
```
KeranYang, line 6: It's recommended by IntelliJ to use the camel case to define constants.

pkg/function/server/server.go
```diff
@@ -27,107 +28,59 @@ func New() *server {
...
-//
-//	func handle(ctx context.Context, keys []string, data functionsdk.Datum) functionsdk.Messages {
-//		_ = data.EventTime() // Event time is available
-//		_ = data.Watermark() // Watermark is available
-//		return functionsdk.MessagesBuilder().Append(functionsdk.NewMessage(data.Value()))
-//	}
-//
-//	func main() {
-//		server.New().RegisterMapper(functionsdk.MapFunc(handle)).Start(context.Background())
-//	}
```
KeranYang, line 40: I am removing this comment because it's a duplicate with what we have in doc.go . Would like to maintain a single source of truth regarding how to use map, which is the flat map example. Same for below

dpadhiar (approved):

pkg/function/server/server.go
```diff
@@ -27,107 +28,59 @@ func New() *server {
...
+// See an example at pkg/function/examples/flatmap/main.go
```
vigith, line 31: i think these examples are based on Godoc guidelines so it will be easy for a casual user reading the doc in the pkg.go.dev.

we should add the example path though so they can dig further.

KeranYang: Right. I see these examples available on pkg.go.dev at https://pkg.go.dev/github.com/numaproj/numaflow-go/pkg/function@v0.4.5, which are generated by the `pkg/function/doc.go` instead of this `server.go`. Hence, we might not need it in `server.go`.

KeranYang: Sorry, I forgot to publish my [comments](https://github.com/numaproj/numaflow-go/pull/68#discussion_r1287701227) after publishing the PR...

whynowy: There's some misleading here, as a private struct, there's no doc exposed for `server`.

PR #69, opened 2023-08-10
feat: add side input grpc service

kohlisid:


pkg/sideinput/configs.go
```diff
@@ -0,0 +1,11 @@
+package sideinput
+
+const (
+	Protocol = "unix"
+	Addr     = "/var/run/numaflow/udsink.sock"
```
jy4096, line 5: udsource

pkg/sideinput/interface.go
```diff
@@ -0,0 +1,19 @@
+package sideinput
+
+import (
+	"context"
+)
+
+// RetrieveSideInputHandler is the interface of map function implementation.
```
jy4096, line 7: `map function implementation` is it for map function?

jy4096 (commented):
I think we will discuss on the sdk design tonight... let's review this PR after we have a conclusion??

pkg/sideinput/service.go
```diff
@@ -0,0 +1,30 @@
...
+	sideinputpb.UnimplementedUserDefinedSideInputServer
+	Retriever RetrieverSideInput
+}
+
+// IsReady returns true to indicate the gRPC connection is ready.
+func (fs *Service) IsReady(context.Context, *emptypb.Empty) (*sideinputpb.ReadyResponse, error) {
+	return &sideinputpb.ReadyResponse{Ready: true}, nil
+}
+
+// RetrieveSideInput applies the function for each side input retrieval request .
```
jy4096, line 22:
```suggestion
// RetrieveSideInput applies the function for each side input retrieval request.
```

pkg/sideinput/interface.go
```diff
@@ -0,0 +1,18 @@
+package sideinput
+
+import (
+	"context"
+)
+
+type RetrieverSideInput interface {
```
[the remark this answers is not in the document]
whynowy: `SideInputRetriever`?

pkg/sideinput/interface.go
```diff
@@ -0,0 +1,18 @@
...
+	// RetrieveSideInput is the function to process each side-input request.
+	RetrieveSideInput(ctx context.Context) MessageSI
```
whynowy, line 9: `Retrieve()`?

pkg/sideinput/interface.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// RetrieverFunc is a utility type used to convert a RetrieveSideInput function to a RetrieverSideInput.
+type RetrieverFunc func(ctx context.Context) MessageSI
```
whynowy, line 13: `RetrieveFunc`

pkg/sideinput/interface.go
whynowy: Shall we give an error option? What do we expect the implementation to do if there's for sure an error will happen (maybe temporarily, but retry won't work)? In this case, what will be the returned []byte? and do we update the exiting data?

pkg/shared/util.go
```diff
@@ -12,10 +12,13 @@ import (
 	"github.com/numaproj/numaflow-go/pkg/info"
 )
 
-func PrepareServer(sockAddr, filePath string) (net.Listener, error) {
-	serverInfo := &info.ServerInfo{Protocol: info.UDS, Language: info.Go, Version: info.GetSDKVersion()}
-	if err := info.Write(serverInfo, info.WithServerInfoFilePath(filePath)); err != nil {
-		return nil, err
+func PrepareServer(infoFilePath string, sockAddr string) (net.Listener, error) {
+	// If infoFilePath is not empty, write the server info to the file.
+	if infoFilePath != "" {
```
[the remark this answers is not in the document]
kohlisid: For Side input we don't write data to the info server, hence will pass empty here. Could be used later on for similar cases

pkg/sideinput/message_si.go
```diff
@@ -0,0 +1,17 @@
+package sideinput
+
+// MessageSI is used to wrap the data return by UserSideInput function.
+// It contains the data value for the given side input parameter requested.
+type MessageSI struct {
+	value []byte
+}
+
+// NewMessageSI creates a new Message with the given value
+func NewMessageSI(value []byte) MessageSI {
```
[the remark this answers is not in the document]
kohlisid: We use the similar patter for transformer where we call it MessageT

pkg/sideinput/message_si.go
```diff
@@ -0,0 +1,17 @@
+package sideinput
+
+// MessageSI is used to wrap the data return by UserSideInput function.
+// It contains the data value for the given side input parameter requested.
+type MessageSI struct {
```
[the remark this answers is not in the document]
kohlisid: Switched to SideInputMessage

pkg/sideinput/interface.go
kohlisid: Have added the error option to the function. In the manager side if there is an error here, then right now we just log it
```
	f := func() {
		if err := sim.execute(ctx, sideInputClient, siStore); err != nil {
			log.Errorw("Failed to execute the call to fetch Side Inputs.", zap.Error(err))
		}
	}
```

pkg/shared/util.go
```diff
@@ -12,10 +12,13 @@ import (
 	"github.com/numaproj/numaflow-go/pkg/info"
 )
 
-func PrepareServer(sockAddr, filePath string) (net.Listener, error) {
-	serverInfo := &info.ServerInfo{Protocol: info.UDS, Language: info.Go, Version: info.GetSDKVersion()}
-	if err := info.Write(serverInfo, info.WithServerInfoFilePath(filePath)); err != nil {
-		return nil, err
+func PrepareServer(infoFilePath string, sockAddr string) (net.Listener, error) {
+	// If infoFilePath is not empty, write the server info to the file.
```
yhl25, line 16: why did we change the order of the arguments?  it will break other servers

pkg/mapper/server.go
```diff
@@ -38,7 +38,7 @@ func (m *server) Start(ctx context.Context) error {
 
 	// write server info to the file
 	// start listening on unix domain socket
-	lis, err := shared.PrepareServer(m.opts.sockAddr, m.opts.serverInfoFilePath)
+	lis, err := shared.PrepareServer(m.opts.serverInfoFilePath, m.opts.sockAddr)
 	if err != nil {
```
yhl25, line 42: revert

pkg/shared/const.go
```diff
@@ -9,6 +9,7 @@ const (
 	SourceAddr            = "/var/run/numaflow/source.sock"
 	SinkAddr              = "/var/run/numaflow/sink.sock"
 	SourceTransformerAddr = "/var/run/numaflow/sourcetransform.sock"
+	SideInputAddr         = "/var/run/numaflow/udsideinput.sock"
```
yhl25, line 12:
```suggestion
	SideInputAddr         = "/var/run/numaflow/sideinput.sock"
```

pkg/sideinput/message_si.go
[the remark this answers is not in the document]
yhl25: just `Message` should be fine, since its inside the sideinput package. WDYT?

pkg/sideinput/message_si.go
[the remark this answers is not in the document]
yhl25: we don't use `MessageT` anymore

pkg/sideinput/server_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	var retrieveHandler = RetrieverFunc(func(ctx context.Context) MessageSI {
+		return NewMessageSI([]byte("test"))
+	})
+	// note: using actual UDS connection
+	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Second)
+	go func() {
+		time.Sleep(3 * time.Second)
+		cancel()
+	}()
```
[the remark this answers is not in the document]
yhl25: +1

yhl25 (approved):
LGTM! Please fix @vigith comments before merging

pkg/sideinput/examples/simple-sideinput/Makefile
```diff
@@ -4,7 +4,7 @@ build:
 
 .PHONY: image
 image: build
-	docker build -t "quay.io/numaio/numaflow-go/sideinput:v3" --target sideinput .
+	docker buildx build -t "quay.io/numaio/numaflow-go/sideinput:latest" --platform linux/amd64,linux/arm64 --target sideinput . --push
```
yhl25, line 7: use v0.5.0 tag instead of latest?

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
```diff
@@ -0,0 +1,46 @@
...
+          rpu: 1
+          duration: 1s
+    - name: si-log
+      udf:
+        container:
+          image: "quay.io/numaio/numaflow-go/udf-sideinput:v0.5.0"
+          imagePullPolicy: Always
+          env:
+            - name: SIDEINPUT_PATH
+              value: "/var/numaflow/side-inputs"
```
jy4096, line 29: so the default value would be `"/var/numaflow/side-inputs"` if not given correct?

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
```diff
@@ -0,0 +1,46 @@
...
+            - name: SIDEINPUT_NAME
+              value: "myticker"
```
jy4096, line 31: this also.. default name will be the same as
```
      sideInputs:
        - myticker
 ```

correct?... maybe add some comments?...

File: pkg/sideinput/interface.go [hunk 17]
```diff
@@ -0,0 +1,19 @@
+package sideinput
+
+import (
+	"context"
+)
+
+// SideInputRetriever is the interface for side input retrieval implementation.
+type SideInputRetriever interface {
+	// RetrieveSideInput is the function to process each side-input request.
+	RetrieveSideInput(ctx context.Context) (Message, error)
```
whynowy, line 10: return `(*Message, error)`?

vigith: let's not return error. what is numaflow going to do if we see an error?

kohlisid: We would just log it right now!
```
if err != nil {
return fmt.Errorf("failed to retrieve side input: %w", err)
}
```

kohlisid: Also @whynowy `(*Message, error)`
We use `Map(ctx context.Context, keys []string, datum Datum) Messages` for our other functions as well. Any specific reason for this change?

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
kohlisid: You mean just add `/path to file directory` instead?

pkg/sideinput/interface.go
vigith: > let's not return error. what is numaflow going to do if we see an error?

user's might expect error handling, which includes retries etc. it might be better to remove error.

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
vigith: doesn't numa export `SIDEINPUT_PATH` ? if not, let's do that in the controller? @whynowy WDYT?

pkg/sideinput/interface.go
whynowy: It's tricky without having an error returned. 

Assume use do error handling - if he/she logs the error without retry (or after several retry), he/she has to return something, and we need to use the returned value to tell if there's any data returned. Or the user has to keep retrying until the gPRC timeouts, which is not ideal.

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
whynowy: We don't need it.

pkg/sideinput/interface.go
vigith: we are not returning error in other places, e.g., `map`; or are we?

whynowy: > Also @whynowy `(*Message, error)` We use `Map(ctx context.Context, keys []string, datum Datum) Messages` for our other functions as well. Any specific reason for this change?

This is for returned value.

Other functions do not expect user to return errors, and it's user's responsibility to process the message, or we block. This one is different, imagine if there's not a sidecar container for retrieve the data, but do it in the numa container, what we should do when there's an problem? We might retry several times, if it still fails, we log it and wait for next run, and do not update the data store. Now, with this happening in the ud container, we would expect this logic still exits. If the the interface does not return an error, and expect the user to retry forever, then we have to wait for the gPRC timeout error (which will be long).

kohlisid:
Added the `no_broadcast` field, did some manual testing to confirm that whenever the `NoBroadcastMessage` no file changes are reflected on the UDF side 
Screenshot 2023-08-24 at 12 38 34 PM
Screenshot 2023-08-24 at 12 38 42 PM

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: my-pipeline
+spec:
+  sideInputs:
+    - name: myticker
+      container:
+        image: "quay.io/numaio/numaflow-go/sideinput:v0.5.0"
+        imagePullPolicy: Always
+      trigger:
+        schedule: "*/2 * * * *"
+        # interval: 180s
```
whynowy, line 13: Remove this line, we do not support `interval` any more.

pkg/sideinput/examples/simple-sideinput/pipeline.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: my-pipeline
+spec:
+  sideInputs:
+    - name: myticker
+      container:
+        image: "quay.io/numaio/numaflow-go/sideinput:v0.5.0"
```
whynowy, line 9: Is this the real image? use something like `udf-sideinput-example` as described in the Dockerfile?

pkg/sideinput/message.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// NoBroadcastMessage creates a new Message with noBroadcast flag set to true
+// This is used to drop the message and not to broadcast it to other side input vertices.
+func NoBroadcastMessage() Message {
+	return Message{value: []byte{}, noBroadcast: true}
+}
+
+// Value returns message value
+func (m Message) Value() []byte {
```
whynowy, line 23: These 2 functions are never used, delete them?

pkg/sideinput/examples/simple-sideinput/udf/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"log"
+	"os"
+	"path"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/numaproj/numaflow-go/pkg/mapper"
+)
+
+var sideInputPath = "/var/numaflow/side-inputs"
```
vigith, line 14: do they have to know this path, can't they get it from ENV var?

pkg/sideinput/examples/simple-sideinput/udf/main.go
```diff
@@ -0,0 +1,73 @@
...
+var sideInputName = "myticker"
+
+func mapFn(_ context.Context, _ []string, _ mapper.Datum) mapper.Messages {
+	// Perform map operation here
+	return mapper.MessagesBuilder().Append(mapper.NewMessage([]byte("test_value")))
```
vigith, line 19:
```suggestion
	return mapper.MessagesBuilder().Append(mapper.NewMessage([]byte("some value")))
```

word `test` has a bad connotation :)

vigith (commented):
LGTM, barring Derek's comments.

pkg/sideinput/examples/simple-sideinput/udf/main.go
kohlisid: We don't expose it as discussed with @whynowy yesterday!

whynowy: Exposing thru ENV might be better, even though they need to memorize something anyways, but it gives us flexibility to have a chance to change the path in the future.

vigith: or wrap in SDK, `GetSideInputDir()`?

kohlisid: Added to shared sdk as `shared.SideInputDir`

whynowy (approved):

PR #72, opened 2023-08-15
feat: add gRPC services for user-defined source

KeranYang:
* Also add a simple source example to demonstrate how users can use the SDK to implement a source.


pkg/apis/proto/source/v1/udsource.proto
```diff
@@ -7,21 +7,21 @@ import "google/protobuf/empty.proto";
 
 package source.v1;
 
-service UserDefinedSource {
+service Source {
   // Read returns a stream of datum responses.
   // The size of the returned ReadResponse is less than or equal to the num_records specified in ReadRequest.
   // If the request timeout is reached on server side, the returned ReadResponse will contain all the datum responses that have been read (which could be an empty list).
-  rpc Read(ReadRequest) returns (stream ReadResponse);
+  rpc ReadFn(ReadRequest) returns (stream DatumResponse);
```
vigith, line 14:
```suggestion
  rpc ReadFn(ReadRequest) returns (stream ReadResponse);
```

we will not be sharing types as those cases are minimal.

pkg/function/examples/assign_event_time/main.go
```diff
@@ -4,8 +4,9 @@ import (
 	"context"
 	"time"
 
-	functionsdk "github.com/numaproj/numaflow-go/pkg/function"
 	"github.com/numaproj/numaflow-go/pkg/function/server"
+
```
vigith, line 8:
```suggestion
```

pkg/sink/examples/log/main.go
```diff
@@ -4,8 +4,9 @@ import (
 	"context"
 	"fmt"
 
-	sinksdk "github.com/numaproj/numaflow-go/pkg/sink"
 	"github.com/numaproj/numaflow-go/pkg/sink/server"
+
```
vigith, line 8:
```suggestion
```

pkg/sink/server/server.go
```diff
@@ -9,10 +9,12 @@ import (
 	"os/signal"
 	"syscall"
 
-	sinkpb "github.com/numaproj/numaflow-go/pkg/apis/proto/sink/v1"
+	"google.golang.org/grpc"
+
 	"github.com/numaproj/numaflow-go/pkg/info"
 	sinksdk "github.com/numaproj/numaflow-go/pkg/sink"
-	"google.golang.org/grpc"
+
```
vigith, line 16:
```suggestion
```

pkg/source/examples/simple_source/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: build
+build:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -o ./dist/simple-source-example main.go
+
+.PHONY: image
+image: build
+	docker build -t "quay.io/numaio/numaflow-go/source-simple-source:latest" --target simple-source .
```
vigith, line 7:
```suggestion
	docker build -t "quay.io/numaio/numaflow-go/source-simple-source:v1" --target simple-source .
```

so we can version images and that way e2e tests won't break?

pkg/source/examples/simple_source/impl/simple_source.go
```diff
@@ -0,0 +1,121 @@
...
+// Right after reading a message, the function marks the offset as to be acked.
+func (s *SimpleSource) Read(_ context.Context, readRequest sourcesdk.ReadRequest, messageCh chan<- model.Message) {
+	// Handle the timeout specification in the read request.
+	ctx, cancel := context.WithTimeout(context.Background(), readRequest.TimeOut())
+	defer cancel()
+
+	// If we have un-acked data, or if we have read all the data,
+	// we return without reading any new data.
+	if len(s.toAckSet) > 0 || s.readIdx >= s.size {
+		return
```
vigith, line 54:
```suggestion
	// If we have un-acked data, we return without reading any new data.
	if len(s.toAckSet) > 0 {
		return
```

we support only unbounded stream so let's not give users ideas :-)

pkg/source/examples/simple_source/impl/simple_source.go
```diff
@@ -0,0 +1,121 @@
...
+	toAckSet map[int64]struct{}
+	lock     *sync.Mutex
+}
+
+// NewSimpleSource creates a SimpleSource by populating the messages array with random strings
+func NewSimpleSource(size int64) *SimpleSource {
+	messages := make([]string, size)
+	for i := int64(0); i < size; i++ {
+		messages[i] = strconv.FormatInt(i, 10)
+	}
```
vigith, line 27: let's keep giving some random strings rather than fix set because our source supports only unbounded streams.

pkg/source/examples/simple_source/impl/simple_source.go
```diff
@@ -0,0 +1,121 @@
...
+		copyOfOriginalToAckSet[k] = v
+	}
+
+	for _, offset := range offsetsToAck {
+		dk := deserializeOffset(offset.Value())
+		if _, ok := s.toAckSet[dk]; !ok {
+			// One of the input offsets is not in the toAckSet.
+			// Revoke all the previous acknowledgements and panic.
+			s.toAckSet = copyOfOriginalToAckSet
+			panic("offsets to ack do not match the toAckSet")
```
vigith, line 106: let's not add this in example as this can be triggered only if there is a bug in numaflow

pkg/source/examples/simple_source/main.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/numaproj/numaflow-go/pkg/source/examples/simple_source/impl"
+	"github.com/numaproj/numaflow-go/pkg/source/server"
+)
+
+func main() {
+	simpleSource := impl.NewSimpleSource(10)
+	server.New(
+		sourcesdk.PendingFunc(simpleSource.Pending),
+		sourcesdk.ReadFunc(simpleSource.Read),
+		sourcesdk.AckFunc(simpleSource.Ack)).
```
vigith, line 16:
```suggestion
		simpleSource).
```

here the `simpleSource` implements the source interface.

pkg/source/interface.go
vigith, line 1: let's not create multiple inteface but a single one.

```golang
type Source interface {
  Read(...)
  Ack(...)
  Pending(...)
}
```

pkg/source/server/server.go
```diff
@@ -0,0 +1,101 @@
...
+
+type server struct {
+	svc *sourcesdk.Service
+}
+
+// New creates a new server object.
+func New(
+	pendingHandler sourcesdk.PendingHandler,
+	readHandler sourcesdk.ReadHandler,
+	ackHandler sourcesdk.AckHandler,
```
vigith, line 27:
```suggestion
	udSource sourcesdk.Source,
```

pass a that implements the interface

pkg/source/service.go
```diff
@@ -0,0 +1,123 @@
...
+	sourcepb "github.com/numaproj/numaflow-go/pkg/apis/proto/source/v1"
+	"github.com/numaproj/numaflow-go/pkg/source/model"
+)
+
+// Service implements the proto gen server interface
+type Service struct {
+	sourcepb.UnimplementedSourceServer
+	PendingHandler PendingHandler
+	ReadHandler    ReadHandler
+	AckHandler     AckHandler
```
vigith, line 19: we do not need this.

pkg/source/examples/simple_source/impl/simple_source.go
```diff
@@ -0,0 +1,88 @@
...
+	// Handle the timeout specification in the read request.
+	ctx, cancel := context.WithTimeout(context.Background(), readRequest.TimeOut())
+	defer cancel()
+
+	// If we have un-acked data, we return without reading any new data.
+	// TODO - we should have a better way to handle this.
+	// In real world, there can be the case when the numa container restarted before acknowledging a batch,
+	// leaving the toAckSet not empty on the UDSource container side.
+	// In this case, for the next batch read, we should read the data from the last acked offset instead of returning.
+	// Our built-in Kafka source follows this logic.
```
KeranYang, line 47: @vigith Thanks for reviewing the PR. I have addressed all of your comments. One more change is that I added this TODO here to tackle the numa container restart case. I will address it in a separate PR.

pkg/source/service.go
```diff
@@ -0,0 +1,121 @@
...
+		timeout: time.Duration(d.Request.GetTimeoutInMs()) * time.Millisecond,
+	}
+	ctx := stream.Context()
+	messageCh := make(chan model.Message)
+
+	done := make(chan bool)
+	go func() {
+		fs.Source.Read(ctx, &request, messageCh)
+		done <- true
+	}()
```
vigith, line 59:
```suggestion
	wg.Add(1)
	go func() {
		defer wg.Done()
		fs.Source.Read(ctx, &request, messageCh)
		close(messageCh)
	}()
```

pkg/source/service.go
```diff
@@ -0,0 +1,121 @@
...
+	finished := false
+	for {
```
vigith, line 61:
```suggestion
	for message := range messageCh {
```

pkg/source/service.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	ctx := stream.Context()
+	messageCh := make(chan model.Message)
+
+	done := make(chan bool)
+	go func() {
+		fs.Source.Read(ctx, &request, messageCh)
+		done <- true
+	}()
+	finished := false
```
vigith, line 60:
```suggestion
```

pkg/source/service.go
```diff
@@ -0,0 +1,121 @@
...
+	for {
+		select {
+		case <-done:
+			finished = true
+		case message, ok := <-messageCh:
+			if !ok {
+				// Channel already closed, not closing again.
+				return nil
+			}
```
vigith, line 69:
```suggestion
```

we won't need the select now

pkg/source/service.go
```diff
@@ -0,0 +1,121 @@
...
+			if err != nil {
+				// The channel may or may not be closed, as we are not sure, leave it to GC.
+				return err
+			}
+		default:
+			if finished {
+				close(messageCh)
+				return nil
+			}
+		}
```
vigith, line 93:
```suggestion
```

or the default

vigith (approved):

pkg/source/service.go
```diff
@@ -0,0 +1,108 @@
...
+	request := readRequest{
+		count:   d.Request.GetNumRecords(),
+		timeout: time.Duration(d.Request.GetTimeoutInMs()) * time.Millisecond,
+	}
+	ctx := stream.Context()
+	messageCh := make(chan model.Message)
+
+	// Start the read in a goroutine
+	go func() {
+		defer close(messageCh)
```
whynowy, line 57: Use the pattern in https://github.com/numaproj/numaflow-go/blob/main/pkg/function/service.go#L153-L167, this will solve the problem of the developer getting the channel closed in the `Read()` implementation.

whynowy: Context: We don't expect the `Read()` implementation to close the channel, but it is possible ppl do that, that that will make line 57 panic.

vigith: We shouldn't let users close the channel and if they do that, let it panic. Ownership will get very tricky otherwise. I did meet with Keran, and we both agreed that we should have strict ownership principles. 

We have added a comment saying, please do not close :)

whynowy (approved):

PR #79, opened 2023-08-28
chore: avoid exposing constants

yhl25:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/numaproj/numaflow-go
 go 1.20
 
 require (
+	github.com/fsnotify/fsnotify v1.6.0
```
vigith, line 6: perhaps we should use go.mod for examples?

whynowy: +1

yhl25: done

pkg/mapstreamer/examples/flatmap_stream/go.mod
```diff
@@ -0,0 +1,15 @@
+module even_odd
+
+go 1.20
+
+require github.com/numaproj/numaflow-go v0.4.6-0.20230828035951-6f79b632ecfe
```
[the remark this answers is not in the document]
yhl25: Following our offline discussion, we have decided against using the relative path for two reasons:

- The examples to consistently point to the latest release version. This approach will help avoid any confusion for users who are trying to understand the examples in the context of the most recent version.
- The Docker file can become quite complex and confusing if we use relative paths.

pkg/sideinput/examples/simple-sideinput/udf/main.go
```diff
@@ -28,7 +28,7 @@ func main() {
 	defer watcher.Close()
 
 	// Add a path to the watcher
-	err = watcher.Add(shared.SideInputDir)
+	err = watcher.Add(sideinput.DirPath)
```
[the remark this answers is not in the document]
yhl25: Since it's just a path, I believe using a constant should be good enough for now. If there is a requirement we can revisit it later

vigith (approved):

PR #81, opened 2023-09-12
fix: allow returning negative value from source pending api

KeranYang:
Such that we enable user to handle any failure scenarios within the pending api execution and to indicate pending is not available by returning a negative number.


pkg/sourcer/interface.go
```diff
@@ -5,15 +5,16 @@ import (
...
 type Sourcer interface {
 	// Read reads the data from the source and sends the data to the message channel.
 	// Read should never attempt to close the message channel as the caller owns the channel.
 	Read(ctx context.Context, readRequest ReadRequest, messageCh chan<- Message)
 	// Ack acknowledges the data from the source.
 	Ack(ctx context.Context, request AckRequest)
 	// Pending returns the number of pending messages.
-	Pending(ctx context.Context) uint64
+	//When the return value is negative, it indicates the pending information is not available.
+	Pending(ctx context.Context) int64
```
juliev0, line 17: comment needs a space after the slashes for consistency

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,27 @@
+name: Go
```
KeranYang, line 1: Not sure why this doesn't kick in with the current PR. Will verify when we have our next PR.

juliev0 (approved):

PR #83, opened 2023-10-02
chore: get ready for 0.5.1 release

KeranYang:
Fix some docs.


pkg/sourcer/doc.go
```diff
@@ -1,2 +1,104 @@
 // Package sourcer implements the server code for User Defined Sourcer in golang.
+//
+// Example:
```
vigith, line 3: you can just make them point to the source code in the examples dir, the code is bound to diverge.

vigith (approved):

pkg/sourcer/doc.go
KeranYang: Agreed. Will do.

PR #84, opened 2023-10-02
chore: update the description for Pending() interface

KeranYang:


pkg/sourcer/interface.go
```diff
@@ -8,12 +8,15 @@ import (
...
 	// Read reads the data from the source and sends the data to the message channel.
+	// If the read request is timed out, the function returns without reading new data.
+	// Right after reading a message, the function marks the offset as to be acked.
 	// Read should never attempt to close the message channel as the caller owns the channel.
 	Read(ctx context.Context, readRequest ReadRequest, messageCh chan<- Message)
 	// Ack acknowledges the data from the source.
 	Ack(ctx context.Context, request AckRequest)
 	// Pending returns the number of pending messages.
 	// When the return value is negative, it indicates the pending information is not available.
+	// With pending information being not available, the Numaflow platform doesn't auto-scale the source.
```
whynowy, line 19: What should be the returned value for `being not available`?

KeranYang: Sorry the line above mentioned that if the returned value is negative, it indicate pending is not available

whynowy: oh.. missed that.

Shall we expose a constant for it?

KeranYang: I don't have a strong opinion on this one. I think asking the user to return a random negative number is already quite straightforward.

whynowy (approved):

PR #90, opened 2023-12-04
fix: require specifying event time when dropping a message using source transformer

KeranYang:


pkg/sourcer/examples/simple_source/go.mod
```diff
@@ -3,14 +3,21 @@ module github.com/numaproj/numaflow-go/pkg/sourcer/examples/simple_source
 go 1.20
 
 require (
-	github.com/numaproj/numaflow-go v0.5.2
-	github.com/stretchr/testify v1.8.4
+	github.com/numaproj/numaflow-go v0.5.1
+	github.com/stretchr/testify v1.8.1
```
vigith, line 7: are we going back?

vigith: i think git showed me stale info

vigith (approved):

PR #91, opened 2023-12-06
chore: adding partitions for user defined source

yhl25:
related to https://github.com/numaproj/numaflow/issues/1405


vigith (approved):

pkg/sourcer/message.go
```diff
@@ -43,11 +43,11 @@ func (m Message) EventTime() time.Time {
 
 type Offset struct {
 	value       []byte
-	partitionId string
+	partitionId int32
```
KeranYang, line 46: I was between `string` and `int` when I worked on the SDK. The question is "Are all user-defined sources using integers to represent partitions?" If the answer is yes, `int32` is perfect. If no, `string` would make more sense because either string or integer can be easily converted to string. If partitions are represented in string format, converting to a unique integer might require some conversion algorithms.

That's also why I named it partitonId instead of partitionIndex.

If we are sure there is no string representation for partitions. Let's go with int32. Otherwise, consider keeping it as string.

yhl25: if the `partitionId` is not a valid number, we are using `defaultPartitionId` (0) in Numaflow.

KeranYang (approved):

pkg/sourcer/message.go
KeranYang: Synced offline. if there is a string representation of a partition, let's give the users full control of how they want to hash it to a unique integer.

pkg/sourcer/interface.go
```diff
@@ -18,6 +18,8 @@ type Sourcer interface {
 	// When the return value is negative, it indicates the pending information is not available.
 	// With pending information being not available, the Numaflow platform doesn't auto-scale the source.
 	Pending(ctx context.Context) int64
+	//Partitions returns the partitions associated with the source.
```
whynowy, line 21: More comments are appreciated. It's really hard for someone who wants to write a UDSource to understand how he/she should implement this function.

vigith (changes requested):

pkg/sourcer/interface.go
vigith: +1

vigith (approved):

PR #102, opened 2024-03-06
Implement update script and build/push workflow

ayildirim21:
Improve the CI of SDK as per  [#1363](https://github.com/numaproj/numaflow/issues/1363)


update_examples.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+function show_help () {
+    echo "Usage: $0 [-h|--help] (-b|--build | -c|--commit-sha <commit_sha>)"
+    echo "  -h, --help         Display help message and exit"
+    echo "  -b, --build        Build the docker images of all the examples"
+    echo "  -c, --commit-sha   Update the numaflow-go version to the specified SHA"
```
KeranYang, line 7:
```suggestion
    echo "  -c, --commit-sha   Update all the examples to depend on the numaflow-go version with the specified SHA"
```

update_examples.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
KeranYang, line 1: nice! Consider adding to the repo a `development.md` file to introduce the script to sdk developers, describing how and when to use it. example: https://github.com/numaproj/numaflow/blob/main/docs/development/development.md

KeranYang (commented):
Thanks for making the change. I left some comments. Also, how was this change tested? and can we fix the snyk security scan failure?

vigith:
you can set it to v0.0.0

```
replace github.com/numaproj/numaflow-go => ../../../../../numaflow-go

require github.com/numaproj/numaflow-go v0.0.0
``

File: .github/workflows/buid-push.yaml [hunk 3]
```diff
@@ -0,0 +1,45 @@
+on:
+  pull_request:
+    types:
+      - closed
+    paths:
+      - "pkg/*/examples/go.sum"
+      - "pkg/*/examples/go.mod"
```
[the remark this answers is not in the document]
ayildirim21: I was thinking that we only build-push when dependency file changes are made, as if we build and push when an SDK change is made it will have no affect on the example images. So my idea was that we would encourage developers to make their SDK changes create a PR, then run the script to update dependency files, and then once the dependency files have been merged we can build and push the new images with the SDK changes.

ayildirim21:
> you can set it to v0.0.0
> 
> ```
> replace github.com/numaproj/numaflow-go => ../../../../../numaflow-go
> 
> require github.com/numaproj/numaflow-go v0.0.0
> ```

If we set to v0.0.0 then in the logs when we print out server info the version that is shown will also be v0.0.0. Would it be better if developers use the script to update the version when they want so it is accurately reflected in the logs?

.github/workflows/buid-push.yaml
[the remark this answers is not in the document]
ayildirim21: I meant as in if we make changes to only the SDK code, it will not have an affect on the images until we update the dependency files.

ayildirim21:
> Thanks for making the change. I left some comments. Also, how was this change tested? and can we fix the synk security scan failure?

I dont have access to snyk so cannot tell why the failures are occuring, but im pretty sure it has something to do with the fact that we modified the go.mod/go.sum files

vigith:
> If we set to v0.0.0 then in the logs when we print out server info the version that is shown will also be v0.0.0. Would it be better if developers use the script to update the version when they want so it is accurately reflected in the logs?

ah, server-info will have the wrong version. good point :)

development.md
```diff
@@ -0,0 +1,33 @@
...
+### Testing
+
+The SDK uses local references in the `go.mod` file, i.e. the `github.com/numaproj/numaflow-go` dependency is pointing to your local
+`numaflow-go` directory. Thus, you can automatically test your SDK changes without the need to push to your forked repository or modify the `go.mod` file.
+Simply make your changes, build and push the desired example image, and you are ready to use it in any pipeline.
+
+### Deploying
+
+1. Create a PR for your changes. Once merged, it will trigger a workflow to build and push the images for all the examples, 
+with the tag `stable`. This consistent tag name is used so that the tags in the E2E test pipelines do not need to be
```
KeranYang: minor: add a link to the numa e2e test folder for reference.

KeranYang (approved):

PR #106, opened 2024-03-12
chore: update development.md

ayildirim21:
update the dev document to include best practice regarding the update script, `update-examples.sh`


development.md
```diff
@@ -31,14 +31,16 @@ This is essentially equivalent to running `make image-push TAG=<tag>` in the exa
 1. Create a PR for your changes. Once merged, it will trigger a workflow to build and push the images for all the examples, 
 with the tag `stable`. This consistent tag name is used so that the tags in the [E2E test pipelines](https://github.com/numaproj/numaflow/tree/main/test) do not need to be 
 updated each time a new change is made. 
-2. After the changes have been merged it is encouraged to update the dependency management files so that the version 
+2. If the change that you have introduced is a breaking change, i.e. adds new features/logic, then after merge it is encouraged to update the dependency management files so that the version
```
KeranYang, line 34:
```suggestion
2. If the change that you have introduced is a breaking change, then after merge you should update the dependency management files so that the version 
```

KeranYang (approved):

update_examples.sh
vigith, line 1: should we move this script `hack/` dir?

ayildirim21: will open a pr to do that

PR #108, opened 2024-03-19
feat: support headers for message

yhl25:


vigith (approved):
document the exposed methods.

Makefile
```diff
@@ -16,4 +16,9 @@ proto:
 generate:
 	go generate ./...
 
+.PHONY: image-push-all
```
KeranYang, line 19: @ayildirim21 please review this makefile target and see if this is something we already worked on as part of the CI improvement task or something we can leverage to simplify our CI script.

pkg/reducer/task_manager.go
```diff
@@ -166,5 +166,5 @@ func generateKey(window *v1.Window, keys []string) string {
 }
 
 func buildDatum(request *v1.ReduceRequest) Datum {
-	return NewHandlerDatum(request.Payload.GetValue(), request.Payload.EventTime.AsTime(), request.Payload.Watermark.AsTime())
+	return NewHandlerDatum(request.Payload.GetValue(), request.Payload.EventTime.AsTime(), request.Payload.Watermark.AsTime(), request.Payload.Headers)
```
KeranYang, line 169: GetEventTime, GetWatermark, GetHeaders - same for below

Makefile
ayildirim21: Can confrim that the existing script already implements this functionality, just need to run:

`./hack/update_examples.sh -bpe`

Will open a PR to remove this makefile target.

vigith: please do not remove the target, instead let the target run `./hack/update_examples -bp`

PR #109, opened 2024-03-21
chore: remove duplicate functionality

ayildirim21:
This make target functionality is already implemented in the update script, can achieve the same result running:

`./hack/update_examples -bp -t test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,3 @@ proto:
 .PHONY: generate
 generate:
 	go generate ./...
-
-.PHONY: image-push-all
-image-push-all:
-	for dir in $(shell find ./pkg -name 'Makefile' -exec dirname {} \;); do \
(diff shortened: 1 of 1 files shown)


KeranYang (approved):

Makefile
```diff
@@ -15,10 +15,3 @@ proto:
 .PHONY: generate
 generate:
 	go generate ./...
-
-.PHONY: image-push-all
-image-push-all:
```
vigith, line 20: don't delete the make target, put `./hack/update_examples -bp -t test` command in the target?

vigith: it is a handy command to have in make

KeranYang: +1. @ayildirim21 please also put a comment on the target to mention where we are pushing to(quay.io).

PR #111, opened 2024-03-25
chore: session sum example

### LINKED ISSUE #110: Missing session-sum source code
session sum is [used](https://github.com/numaproj/numaflow/blob/main/test/reduce-two-e2e/testdata/session-reduce/simple-session-sum-pipeline.yaml#L19) in our numaflow e2e. Let's add it to numaflow-go examples so that we have it source code controlled.

yhl25:
fixes #110


pkg/reducestreamer/task_manager.go
```diff
@@ -141,26 +139,22 @@ func (rtm *reduceStreamTaskManager) OutputChannel() <-chan *v1.ReduceResponse {
...
-	}
-
-	for _, task := range tasks {
 		<-task.doneCh
+		if eofResponse == nil {
+			eofResponse = task.buildEOFResponse()
+		}
 	}
+	rtm.responseCh <- eofResponse
+
```
KeranYang, line 150: Is this a behaviour change? Before, each task, if not merged, sends its own EOFResponse to responseCh. After, when receiving an EOF, we loop through the active tasks and only send the last EOFResponse.

KeranYang: I think we should include verifying EOF in our unit tests. Currently, we are [not](https://github.com/numaproj/numaflow-go/blob/main/pkg/sessionreducer/service_test.go#L1472).

KeranYang: I am approving as this is sort of urgent. We can address the unit test comment as a separate issue: https://github.com/numaproj/numaflow-go/issues/113

KeranYang (approved):

PR #115, opened 2024-04-09
chore: add minimum client version to server info struct

ayildirim21:
related to: [#1504](https://github.com/numaproj/numaflow/issues/1504)

### DIFF
--- a/pkg/info/types.go
+++ b/pkg/info/types.go
@@ -15,10 +15,13 @@ const (
 	Java   Language = "java"
 )
 
+const MinimumClientVersion = ">= 1.2.0-0"
+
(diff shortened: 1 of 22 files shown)


KeranYang (approved):

pkg/info/types.go
```diff
@@ -15,10 +15,13 @@ const (
 	Java   Language = "java"
 )
 
+const MinimumClientVersion = ">= 1.2.0-0"
```
vigith, line 18: let's create a `release.md` file and document that this field should be updated if there is a change?

vigith (approved):

PR #120, opened 2024-04-15
feat: Fallback Sink

yhl25:


vigith (approved):

pkg/sinker/types.go
```diff
@@ -8,6 +8,8 @@ type Response struct {
 	Success bool `json:"success"`
 	// Err represents the error message when "success" is false.
 	Err string `json:"err,omitempty"`
+	// Fallback is true if the message to be sent to the fallback sink.
+	Fallback bool `json:"fallback,omitempty"`
```
whynowy, line 12: Shall we combine `Fallback` and `Success` to use a new property?

yhl25: I thought of using an enum, but enums might get tricky in some languages so went with two separate boolean values.

PR #122, opened 2024-04-18
chore: fallback log sink example

yhl25:


KeranYang (commented):
icydk: https://github.com/numaproj/numaflow-go/blob/main/development.md#adding-a-new-example

KeranYang (approved):

PR #128, opened 2024-04-24
chore: UDS based multiproc

kohlisid:


pkg/info/server_info_test.go
```diff
@@ -55,7 +55,7 @@ func Test_Read_Write(t *testing.T) {
 	filepath := os.TempDir() + "/server-info"
 	defer os.Remove(filepath)
 	info := &ServerInfo{
-		Protocol:               TCP,
+		MultiProcServer:        true,
```
whynowy, line 58: Instead of updating the struct, can we keep the `protocol`, and use the metadata to populate the info? For example, in the metadata, `multiProc: 4` means multProcServer with 4.

kohlisid:
Closing this PR as it is no longer required.

PR #132, opened 2024-06-28
feat: implement batch map

kohlisid:


pkg/apis/proto/map/v1/map.proto
```diff
@@ -13,6 +13,14 @@ service Map {
...
   // IsReady is the heartbeat endpoint for gRPC.
   rpc IsReady(google.protobuf.Empty) returns (ReadyResponse);
+
+  // BatchMapFn is a bi-directional streaming rpc which applies a
+  // batchMap function on each element of the stream and then returns streams
+  // back MapResponse elements.
+  // TODO(map-batch): in the target state when we move the current
+  // unary implementation to bi-di as well, we can rename this and
+  // use a single rpc for both.
+  rpc BatchMapFn(stream MapRequest) returns (stream MapResponse);
```
vigith, line 23:
```suggestion
  rpc MapStreamFn(stream MapRequest) returns (stream MapResponse);
```

pkg/mapper/server.go
```diff
@@ -31,6 +34,21 @@ func NewServer(m Mapper, inputOptions ...Option) numaflow.Server {
 	return s
 }
 
+// NewBatchMapServer creates a new batch map server.
+// TODO(map-batch): as this would be a streaming server should we see if there are some options (like maxMessageSize)
+// which are different than unary server which are optimal for this use case.
+func NewBatchMapServer(m BatchMapper, inputOptions ...Option) numaflow.Server {
```
vigith, line 40:
```suggestion
func NewBatchServer(m BatchMapper, inputOptions ...Option) numaflow.Server {
```

pkg/mapper/examples/batchmap/main.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func mapFn(_ context.Context, datums []mapper.Datum) mapper.BatchResponses {
+	batchResponses := mapper.BatchResponsesBuilder()
+	log.Println("MYDEBUG: length of input ", len(datums))
+	for _, d := range datums {
+		msg := d.Value()
+		_ = d.EventTime() // Event time is available
+		_ = d.Watermark() // Watermark is available
+		results := mapper.NewBatchResponse(d.Id())
```
magelisk, line 17: Is an empty response valid to ensure prop ACK'ing of message without necessarily assigning a response message to it? 

Ex. my batch is a list of JSON, but in my handler I take combine them into a list of JSON (psedo-reduce). One of the `NewBatchResponse` would align with the new message, but the rest I would want to acknowledge as complete, but they wouldn't have an actual response.

kohlisid: We would still need to add an empty batch response for those IDs with an empty list. We expect that the number of requests in batch and number of responses to be equal. 
But for you use case if you pass on an empty response with the correct ID, those should be acked 
`
results := mapper.NewBatchResponse(d.Id())
batchResponses = batchResponses.Append(results)
`

vigith:
merging to a tmp branch

PR #133, opened 2024-07-02
feat: implement batch map

kohlisid:
Fixes [#1786](https://github.com/numaproj/numaflow/issues/1786)


pkg/mapper/service.go
```diff
@@ -13,13 +16,15 @@ const (
 	address               = "/var/run/numaflow/map.sock"
 	defaultMaxMessageSize = 1024 * 1024 * 64
 	serverInfoFilePath    = "/var/run/numaflow/mapper-server-info"
+	emptyId               = ""
```
vigith, line 19: what is this?

pkg/batchmapper/examples/batchmap-flatmap/go.mod
```diff
@@ -0,0 +1,17 @@
+module flatmap
+
+go 1.20
+
+replace github.com/numaproj/numaflow-go => ../../../..
+
+require github.com/numaproj/numaflow-go v0.7.0-rc2
```
vigith, line 7: fix this?

pkg/batchmapper/examples/batchmap-flatmap/go.mod
```diff
@@ -0,0 +1,17 @@
+module flatmap
+
+go 1.20
```
vigith, line 3:
```suggestion
go 1.22
```

vigith (approved):
LGTM

pkg/batchmapper/interface.go
```diff
@@ -0,0 +1,37 @@
...
+	Headers() map[string]string
+	// Id returns the unique ID set for the given message
+	Id() string
+	// Keys returns the keys associated with a given datum
+	Keys() []string
+}
+
+// BatchMapper is the interface for a Batch Map mode where the user is given a list
+// of messages, and they return the consolidated response for all of them together.
+type BatchMapper interface {
```
whynowy, line 26: Naming convention - we have `MapStreamer`.

pkg/batchmapper/interface.go
```diff
@@ -0,0 +1,37 @@
...
+	// BatchMap is the function which processes a list of input messages
+	BatchMap(ctx context.Context, datums []Datum) BatchResponses
```
whynowy, line 28: Naming convention - we have `MapStream()`

pkg/batchmapper/examples/batchmap-flatmap/main.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"context"
+	"log"
+	"strings"
+
+	"github.com/numaproj/numaflow-go/pkg/batchmapper"
+)
+
+func batchMapFn(_ context.Context, datums []batchmapper.Datum) batchmapper.BatchResponses {
```
whynowy, line 11: Can we do a channel input here? This will be consistent as the sinker API, and performance wise, it's slight better.

`batchMapFn(_ context.Context, datums <-chan batchmapper.Datum) batchmapper.BatchResponses {`

pkg/batchmapper/interface.go
whynowy: This makes more sense, we should rename existing `MapStreamer` to `StreamMapper`, similarly, `ReduceStreamer` -> `StreamReducer`.

pkg/batchmapper/examples/batchmap-flatmap/main.go
kohlisid: @yhl25 wdyt?

pkg/batchmapper/interface.go
vigith: > This makes more sense, we should rename existing
why do you think this makes more sense, just curious :)

kohlisid: We discussed and felt that having the operation (ie Map, Reduce, Transform) -> to be the action here.
So keeping mapper, reducer, transformer would be consistent. And the the prefix can be the special type if any.
StreamMapper, BatchMapper, StreamReducer, SessionReducer

pkg/batchmapper/examples/batchmap-flatmap/main.go
kohlisid: From my side I think 
1) From a efficiency stand point, I guess having a channel is better as it might help not buffer the input requests multiple times. As we first buffer on the SDK side and then pass this onto the user which would probably iterate over this as well to read the inputs. 

2) From an understanding and interface side I guess having an array is more closer to the _batch_ philosophy i guess than an iterator. Where you have a concrete list of elements in front of you and can choose to do whatever further with them.

vigith: +1 on consistency though

pkg/apis/proto/batchmap/v1/batchmap.proto
```diff
@@ -0,0 +1,52 @@
...
+  // BatchMapFn is a bi-directional streaming rpc which applies a
+  // Map function on each BatchMapRequest element of the stream and then returns streams
+  // back MapResponse elements.
+  rpc BatchMapFn(stream BatchMapRequest) returns (stream BatchMapResponse);
+}
+
+/**
+ * MapRequest represents a request element.
+ */
+message BatchMapRequest {
```
yhl25, line 23:
```suggestion

/**
 * BatchMapRequest represents a request element.
 */
message BatchMapRequest {
```

pkg/apis/proto/batchmap/v1/batchmap.proto
```diff
@@ -0,0 +1,52 @@
...
+  google.protobuf.Timestamp watermark = 4;
+  map<string, string> headers = 5;
+  // This ID is used uniquely identify a map request
+  string id = 6;
+}
+
+/**
+ * MapResponse represents a response element.
+ */
+message BatchMapResponse {
```
yhl25, line 36:
```suggestion
/**
 * BatchMapResponse represents a response element.
 */
message BatchMapResponse {
```

pkg/batchmapper/examples/batchmap-flatmap/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+####################################################################################################
+# flatmap
+####################################################################################################
+FROM scratch as flatmap
+ARG ARCH
+COPY --from=base /usr/share/zoneinfo /usr/share/zoneinfo
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=base /bin/flatmap-example /bin/flatmap-example
+ENTRYPOINT [ "/bin/flatmap-example" ]
```
yhl25, line 20:
```suggestion
COPY --from=base /bin/batchmap-flatmap /bin/batchmap-flatmap

ENTRYPOINT [ "/bin/batchmap-flatmap
" ]
```

pkg/batchmapper/examples/batchmap-flatmap/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# base
+####################################################################################################
+FROM alpine:3.12.3 as base
+RUN apk update && apk upgrade && \
+    apk add ca-certificates && \
+    apk --no-cache add tzdata
+
+COPY dist/flatmap-example /bin/flatmap-example
+RUN chmod +x /bin/flatmap-example
+
```
yhl25, line 11: same here

pkg/batchmapper/examples/batchmap-flatmap/Makefile
```diff
@@ -0,0 +1,18 @@
+TAG ?= stable
+PUSH ?= false
+
+.PHONY: build
+build:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -o ./dist/flatmap-example main.go
+
```
yhl25, line 7: same here

pkg/batchmapper/examples/batchmap-flatmap/main.go
yhl25: I was looking more inclined towards the array approach because of simplicity, but since we are using a channel in the sink lets go with the same here as well to be consistent.

pkg/batchmapper/interface.go
yhl25: MapBatcher doesn't make much sense to me :)

pkg/batchmapper/service.go
```diff
@@ -0,0 +1,90 @@
...
+	// Apply the user BatchMap implementation function
+	responses := fs.BatchMapper.BatchMap(ctx, datums)
+
+	// If the number of responses received does not align with the request batch size,
+	// we will not be able to process the data correctly.
+	// This should be marked as an error and shown to the user.
+	if len(responses.Items()) != len(datums) {
+		errMsg := "batchMapFn: mismatch between length of batch requests and responses"
+		log.Panic(errMsg)
+	}
```
yhl25, line 63: is it possible to avoid panic?

pkg/batchmapper/examples/batchmap-flatmap/Makefile
```diff
@@ -0,0 +1,18 @@
...
+.PHONY: image-push
+image-push: build
+	docker buildx build -t "quay.io/numaio/numaflow-go/batch-map-flatmap:${TAG}" --platform linux/amd64,linux/arm64 --target flatmap . --push
```
KeranYang, line 10: Please apply the changes in https://github.com/numaproj/numaflow-go/pull/134 here and add batchmap example to build-push.yaml so that it gets included in auto-update.

pkg/batchmapper/service.go
kohlisid: We can choose either to panic or send an error back(numa will be restarted in both cases), based on an earlier discussion with @vigith we thought that having a panic here would be a clear indication for the user that there is an issue. And as this in a mismatch in the implementation from the user side for not providing equal number of responses, the reason for restarts and error gives better clarity there. 
But this is our decision on how do we want to keep the behavior in such scenarios.
@vigith @whynowy Any preferences?

pkg/batchmapper/interface.go
kohlisid: As discussed will create new issues to rename the previous implementations

pkg/batchmapper/examples/batchmap-flatmap/Makefile
kohlisid: @KeranYang I have added the changes, please verify once

pkg/sourcetransformer/examples/event_time_filter/Makefile
```diff
@@ -1,18 +1,19 @@
 TAG ?= stable
 PUSH ?= false
+IMAGE_REGISTRY = quay.io/numaio/numaflow-go/mapt-event-time-filter:${TAG}
```
KeranYang, line 3: Just curious, if we update the branch to bring in Abdullah's change, should these Makefile changes NOT show up in the diff?

kohlisid: I did pull in his changes into this branch. I'm guessing that should be taken care of while merging in

pkg/batchmapper/service.go
```diff
@@ -0,0 +1,126 @@
...
+			// close the input data channel to indicate that no more messages expected
+			close(datumStreamCh)
+			log.Println("BatchMapFn: Got an error while recv() on stream", err)
+			return err
+		}
+		var hd = NewHandlerDatum(d.GetValue(), d.GetEventTime().AsTime(), d.GetWatermark().AsTime(), d.GetHeaders(), d.GetId(), d.GetKeys())
+		// send the datum to the input channel
+		datumStreamCh <- hd
+		// Increase the counter for number of requests received
+		totalRequests.Inc()
```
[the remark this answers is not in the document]
kohlisid: I preferred to update this once we are sure that the message has been sent to the channel, ie consumed by the user code. So if they return early, at that point we are sure about the correct count of number of requests sent to them.
Any things that come to your mind which can cause an issue?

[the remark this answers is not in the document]
KeranYang: Sorry, I forgot to mention my concern. I was looking at this piece of code:

```
if err != nil {
			// close the input data channel to indicate that no more messages expected
			close(datumStreamCh)
			log.Println("BatchMapFn: Got an error while recv() on stream", err)
			return err
		}
```

Since the error is not EOF, I wonder if we still receive a datum. If so, with our current flow, the request count is not increased. Hence we received a datum but didn't get to send it to the channel. On the collecting response side, the count still matches, would it cause a data loss?

We are returning an error, will it fail the entire batch? Please confirm. Thanks!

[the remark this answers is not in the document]
kohlisid: In case of an error we would have return and fail + replay the entire batch

KeranYang (approved):

pkg/batchmapper/service_test.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func TestService_BatchMapFn(t *testing.T) {
+	tests := []struct {
+		name        string
+		handler     BatchMapper
+		input       []*batchmappb.BatchMapRequest
+		expected    []*batchmappb.BatchMapResponse
+		expectedErr bool
+	}{
```
yhl25, line 98: Can we add a test to cover the error/panic case inside the user code?

yhl25 (approved):

pkg/batchmapper/service_test.go
kohlisid: In the current setup, this is difficult to test via UT directly. I tried to add the test there but the assert.Panics (which internally uses recover) was not able to catch it as the panic is invoked from within another goroutine. Would need to research more on how can we cover something like this. In python i'm able to mock the panic and test which wasn't working out here.

I tried few manual scenarios to induce such panics which were working fine.

yhl25 (approved):
LGTM!

PR #139, opened 2024-08-14
chore: directly use hash key to check messages

KeranYang:
* Use the environment variable SINK_HASH_KEY to write to redis sink, which is used by RedisSinkContains() in e2e testing. It's a more generic approach than before.


pkg/sinker/examples/redis-sink/main.go
```diff
@@ -26,14 +26,32 @@ func (rds *redisTestSink) Sink(ctx context.Context, datumStreamCh <-chan sinksdk
...
 		// Each field of a hash is the content of a message and value of the field is the no. of occurrences of the message.
-		hkey := fmt.Sprintf("%s:%s", os.Getenv("NUMAFLOW_PIPELINE_NAME"), os.Getenv("NUMAFLOW_VERTEX_NAME"))
-		err := client.HIncrBy(ctx, hkey, string(d.Value()), 1).Err()
+		var hashKey string
+
+		monoVertexName := os.Getenv("NUMAFLOW_MONO_VERTEX_NAME")
+		pipelineName := os.Getenv("NUMAFLOW_PIPELINE_NAME")
+		vertexName := os.Getenv("NUMAFLOW_VERTEX_NAME")
+
+		if !(monoVertexName != "" || (pipelineName != "" && vertexName != "")) {
```
yhl25, line 36: Instead of generating `hashKey` from the Numaflow platform envs, should we consider passing it as an argument? This would allow us to avoid these conditional checks. I'm okay with introducing a separate environment variable just for the `hashKey`. Using the Numaflow monovertex and pipeline environment variables doesn't seem right, as it restricts us. Ability to configure the `hashKey` would give us much more control. Additionally, sinks should be independent and not depend on internal environment variables.

KeranYang: @yhl25 Thanks for putting in the thoughts. I would prefer to keep it as it is. `hashKey` is a concept that only lives in the world of the Redis sink hence from an ownership perspective, we shouldn't create an environment variable for it from the controller side. We introduced `hashKey` to ensure we have a globally unique key to identify where the message is written from, and using the supported environment variables perfectly serves the purpose.

> Additionally, sinks should be independent and not depend on internal environment variables.

Please elaborate more on this one. I use these environment variables since they are listed in the [udsink page](https://numaflow.numaproj.io/user-guide/sinks/user-defined-sinks/#available-environment-variables), so open to be used.

> Should we consider passing it as an argument?

Technically we can, but it will delegate the key construction to each of the sink specs, which is not as simple as directly use environment variables.

yhl25: "`hashKey` is a concept that only lives in the world of the Redis sink". Hence, we actually shouldn't be using platform environment variables for constructing it. I meant setting `hashKey` at the container level using envs, not introducing a new controller-level env variable. Constructing the `hashKey` itself should not be a significant issue as long as we ensure that it remains unique across tests. We could potentially use the test name itself as the hashKey.

KeranYang: Discussed offline, I will test it out tmr. Thanks, Yashash!

KeranYang: @yhl25 Please re-review. I will update the test cases on the numaflow end accordingly. Thanks!

yhl25 (approved):

PR #148, opened 2024-09-14
chore: update examples to latest sdk version

yhl25:


vigith (commented):
Why didn't we use local path reference in go.mod for examples?

vigith:
> Why didn't we use local path reference in go.mod for examples?

@KeranYang do you remember?

vigith (approved):
Looks like we still have to give version

KeranYang:
@vigith we declare in `go,mod` required version of numaflow-go, but when building the `stable` image, we use the local reference. We don't necessarily need to update the version but I am ok with making it up-to-date. cc: @ayildirim21

PR #151, opened 2024-09-23
feat: Bidirectional Streaming for UDSink

yhl25:


vigith (approved):
please log errs along with returning them.

PR #165, opened 2024-11-05
chore: shutdown when we see non retryable udf errors

yhl25:


pkg/batchmapper/service.go
```diff
@@ -91,19 +91,39 @@ func (fs *Service) performHandshake(stream mappb.Map_MapFnServer) error {
...
 }
 
+// recvWithContext wraps stream.Recv() to respect context cancellation.
+func recvWithContext(ctx context.Context, stream mappb.Map_MapFnServer) (*mappb.MapRequest, error) {
+	type recvResult struct {
+		req *mappb.MapRequest
+		err error
+	}
+
+	resultCh := make(chan recvResult, 1)
```
KeranYang, line 101: do we need this channel and the goroutine if we put `stream.Recv` inside select?

pkg/mapper/service.go
```diff
@@ -72,13 +94,13 @@ func (fs *Service) MapFn(stream mappb.Map_MapFnServer) error {
 	// Read requests from the stream and process them
 outer:
 	for {
-		select {
-		case <-groupCtx.Done():
+		req, err := recvWithContext(groupCtx, stream)
+		if errors.Is(err, context.Canceled) {
```
KeranYang, line 98: I see different behaviour between map and map batchmap above. batchmap is returning the err while map is breaking the outer loop without returning the err. Is this expected? same for map stream.

KeranYang (approved):

vigith (approved):

PR #171, opened 2024-12-19
chore: Map Stream to Support Concurrent Requests

yhl25:


pkg/mapstreamer/service.go
```diff
@@ -64,43 +64,59 @@ func (fs *Service) MapFn(stream mappb.Map_MapFnServer) error {
...
 		if err != nil {
 			log.Printf("Failed to receive request: %v", err)
-			return err
+			readErr = err
+			// read loop is not part of the error group, so we need to cancel the context
+			// to signal the other goroutines to stop processing.
+			cancel()
+			break outer
 		}
+		g.Go(func() error {
```
vigith, line 94: this is bounded because we control the concurrency on the numaflow side, right?

vigith (approved):

pkg/mapstreamer/service.go
yhl25: yes

PR #172, opened 2025-01-16
feat: new Mapper example which serves as a slow container for testing purposes

juliev0:
For Numaplane, it's useful to be able to simulate a slow Vertex. This does a single "cat" but does a sleep first. 

Sleep time is configurable by environment variable.


pkg/mapper/examples/cat_sleep/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"time"
+
+	"github.com/numaproj/numaflow-go/pkg/mapper"
+)
+
+type CatSleep struct {
+}
+
+func (e *CatSleep) Map(ctx context.Context, keys []string, d mapper.Datum) mapper.Messages {
+	time.Sleep(10 * time.Second)
```
KeranYang, line 15: can we make the duration configurable? thanks!

juliev0: done :)

juliev0: Nice, thanks! I will do that. I was actually wondering if I'd need to go in here all the time and update that.

pkg/mapper/examples/cat_sleep/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+const DEFAULT_SLEEP_SECONDS = 10
+
+type CatSleep struct {
+}
+
+func (e *CatSleep) Map(ctx context.Context, keys []string, d mapper.Datum) mapper.Messages {
+
+	// sleep for as long as the environment variable indicates (or default if not configured)
+	sleepSeconds := DEFAULT_SLEEP_SECONDS
```
vigith, line 21: you could move this inside `CatSleep`?

pkg/mapper/examples/cat_sleep/main.go
```diff
@@ -0,0 +1,42 @@
...
+	if secondsString == "" {
+		log.Printf("SLEEP_SECONDS environment variable not set, using default %d seconds\n", DEFAULT_SLEEP_SECONDS)
+	} else {
+		val, err := strconv.Atoi(secondsString)
+		if err != nil {
+			log.Printf("SLEEP_SECONDS environment variable %q not an int, using default %d seconds\n", secondsString, DEFAULT_SLEEP_SECONDS)
+		} else {
+			sleepSeconds = val
+		}
+	}
```
vigith, line 32: this can be done in main?

pkg/mapper/examples/cat_sleep/main.go
```diff
@@ -0,0 +1,42 @@
...
+	time.Sleep(time.Duration(sleepSeconds) * time.Second)
```
vigith, line 33:
```suggestion
	time.Sleep(time.Duration(e.sleepSeconds) * time.Second)
```

vigith (commented):
NIT

pkg/mapper/examples/cat_sleep/main.go
juliev0: True. I suppose I can make this purposely slow container more efficient. :D 

but it would still be neater though, I agree

juliev0:
Decided to rename to 'slow cat' and also relocated the logic to reduce code redundancy (and reduce log spew, which would've been pretty ugly too) (thanks for the suggestion @vigith)

juliev0:
By the way, I've tested the image out in my pipeline, both with and without environment variable specified.

vigith (approved):

KeranYang (approved):

PR #173, opened 2025-01-25
chore: update panic error handling in containers

### LINKED ISSUE #174: Add common prefix and formalize error messages
Streamline the error logging and returned errors in Golang SDK during application exception and shutdown.

adarsh0728:
1. Adds prefix `UDF_EXECUTION_ERROR` in panic mssgs in all containers to use it across SDK's to formalize error from user's code. This will help in debuggability and distinguishing between error in platform and user code.

2. Fixed error handling in case of sideInput, sourceTransformer

3. Error handling In case of reducer, sessionReducer, reduceStreamer (we are not sending error back in case of panics) to be taken up in separate PR.

4. Added debug stack in details part which represents a serialized `google.rpc.Status`, the bytes should be parsed as a Protobuf message at client side.

fixes #174


pkg/sourcetransformer/service.go
```diff
@@ -158,7 +158,7 @@ func (fs *Service) handleRequest(ctx context.Context, req *v1.SourceTransformReq
 	defer func() {
 		if r := recover(); r != nil {
 			log.Printf("panic inside handler: %v %v", r, string(debug.Stack()))
```
KeranYang, line 160: Just curious, did we see this log being printed at transformer container?

pkg/sourcetransformer/service.go
```diff
@@ -158,7 +158,7 @@ func (fs *Service) handleRequest(ctx context.Context, req *v1.SourceTransformReq
...
-			err = errTransformerPanic
+			err = fmt.Errorf("%s: %v", errTransformerPanic, r)
```
KeranYang, line 161: +1

pkg/sourcetransformer/service.go
```diff
@@ -123,12 +123,12 @@ outer:
 	if err := grp.Wait(); err != nil {
 		log.Printf("Stopping the SourceTransformFn with err, %s", err)
 		fs.shutdownCh <- struct{}{}
-		return status.Errorf(codes.Internal, err.Error())
+		return status.Errorf(codes.Internal, "%s", err.Error())
```
KeranYang, line 126: Do we need this change?

KeranYang (commented):
Thanks for making the change Adarsh. Could you fix the unit test?

pkg/sourcetransformer/service.go
KeranYang: Asking because based on the log before this change

```
"task failed: Transformer Error - failed to receive transformer response: status: Internal, message: \"transformer function panicked\"
```

It seems the err.Error gets printed correctly and we don't necessarily need  `"%s"`.

adarsh0728:
> Thanks for making the change Adarsh. Could you fix the unit test?

yes! forgot to convert it into draft :)

pkg/sourcetransformer/service.go
adarsh0728: not a necessary change but I was getting this
`non-constant format string in call to google.golang.org/grpc/status.Errorfprintf[default](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/printf)`

Expecting format string to be constant.

`status.Errorf` internally calls  `fmt.Sprintf`

```
func Errorf(c codes.Code, format string, a ...any) error {
	return Error(c, fmt.Sprintf(format, a...))
}
```

pkg/sourcetransformer/service.go
adarsh0728: yes, mostly! in cases where multiple go routines panic at the same time, then may be log outputs might get mixed

pkg/mapper/service.go
```diff
@@ -23,6 +23,8 @@ const (
 	serverInfoFilePath    = "/var/run/numaflow/mapper-server-info"
 )
 
+var errMapHandlerPanic = errors.New("USER_CODE_ERROR: map handler panicked")
```
vigith, line 26:
```suggestion
var errMapHandlerPanic = errors.New("USER_CODE_ERROR(map)")
```

Isn't this enough?

KeranYang (approved):

yhl25:
Please return the stack trace along with the error message, in future when we want to show errors in the UI, stack traces will help.

pkg/mapper/service.go
adarsh0728: I guess yes, since we are capturing only panics/uncaught exceptions..this should suffice

yhl25 (approved):

PR #175, opened 2025-02-20
fix: server stuck during shutdown in multi-partitioned pipeline

yhl25:


vigith (approved):

KeranYang:
Do we need to apply similar changes to other SDKs?

PR #176, opened 2025-02-22
feat: Custom Serve Store for integration with any user-defined store

vigith:


pkg/batchmapper/interface.go
```diff
@@ -16,7 +16,7 @@ type Datum interface {
 	// Headers returns the headers of the message.
 	Headers() map[string]string
 	// Id returns the unique ID set for the given message
-	Id() string
+	ID() string
```
vigith, line 19: isn't this a breaking change?

yhl25 (approved):

pkg/batchmapper/interface.go
yhl25: reverted the change.

PR #178, opened 2025-02-23
feat: Sink serve API

BulkBeing:


yhl25 (approved):
We might have to support multiple payloads for a id, we can do it in a follow-up PR.

PR #186, opened 2025-03-22
fix: stream sorter example, return when the channel is closed

yhl25:


pkg/accumulator/examples/streamsorter/main.go
```diff
@@ -26,7 +26,11 @@ func (s *streamSorter) Accumulate(ctx context.Context, input <-chan accumulator.
 		case <-ctx.Done():
 			log.Println("Exiting the Accumulator")
 			return
-		case datum := <-input:
+		case datum, ok := <-input:
```
vigith, line 29:
```suggestion
		case datum, ok := <-input:
		// this case happens due to timeout
```

vigith (approved):

PR #188, opened 2025-03-29
fix: redis sink example to not create client for every sink invocation

yhl25:


ayildirim21 (approved):

vigith (approved):

KeranYang:
@yhl25 Thank you for fixing!

PR #204, opened 2025-08-10
fix: client cancellation should not trigger grpc shutdown

yhl25:
### Problem
Client cancellation was triggering server shutdown and udf container restart.

### Fix
If the error type is `context.Cancelled` we don't send shutdown signal to the server.

### DIFF
--- a/pkg/accumulator/service.go
+++ b/pkg/accumulator/service.go
@@ -106,7 +106,7 @@ func (fs *Service) AccumulateFn(stream accumulatorpb.Accumulator_AccumulateFnSer
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/accumulator/service.go
```diff
@@ -108,7 +107,7 @@ func (fs *Service) AccumulateFn(stream accumulatorpb.Accumulator_AccumulateFnSer
 	}
 
 	// wait for all goroutines to finish
```
vigith, line 109:
```suggestion
	// wait for all goroutines to finish, ignore client cancellation errors
```

vigith (approved):

PR #208, opened 2025-09-07
feat: metadata implementation for source, map and sink

adarsh0728:


pkg/sourcer/service.go
```diff
@@ -140,8 +141,9 @@ func (fs *Service) receiveReadRequests(ctx context.Context, stream sourcepb.Sour
 			close(messageCh)
 		}()
 		request := readRequest{
-			count:   req.Request.GetNumRecords(),
-			timeout: time.Duration(req.Request.GetTimeoutInMs()) * time.Millisecond,
+			count:    req.Request.GetNumRecords(),
+			timeout:  time.Duration(req.Request.GetTimeoutInMs()) * time.Millisecond,
+			metadata: fromProto(req.Request.GetMetadata()),
```
yhl25, line 146: Metadata should be originated from source, request doesn't need to expose metadata we can inject the system metadata after the messages are read.

pkg/sourcer/metadata.go
```diff
@@ -0,0 +1,61 @@
...
+	return md.data
+}
+
+// NewMetadata creates a new source Metadata with provided user metadata.
+// If nil is passed, an empty user metadata map is created.
+// Example:
+//
+//	user := make(UserMetadata)
+//	user["kafka-headers"] = map[string][]byte{"k1": []byte("v1")}
+//	sourcer.NewMessage(value, offset, eventTime).WithMetadata(sourcer.NewMetadata(user))
```
yhl25: update the comment to use new semantics?

pkg/mapper/metadata.go
```diff
@@ -0,0 +1,178 @@
...
+// Usage example: userMetadata := datum.UserMetadata()
+// value := userMetadata.Value("group-name", "key")
+func (md UserMetadata) Value(group, key string) []byte {
+	return md.data[group][key]
+}
+
+// SetKVGroup sets a group of key-value pairs under the provided group name.
+// Usage example: userMetadata := NewUserMetadata()
+// userMetadata.SetKVGroup("group-name", map[string][]byte{"key": []byte("value")})
+func (md UserMetadata) SetKVGroup(group string, kv map[string][]byte) {
```
BulkBeing: It's better to use pointer receiver for all methods https://dave.cheney.net/2016/03/19/should-methods-be-declared-on-t-or-t
Value receivers can easily create bugs. Eg
```go
type Metadata struct {
	m map[string]string
}

func(m Metadata) Set(k, v string) {
	if m.m == nil {
		m.m = make(map[string]string)
	}
	m.m[k] = v
}

func main() {
	var m Metadata
	m.Set("a", "b") // mutates the copy
	fmt.Println(m.m) // map[]
}
```

If we later have a need to introduce a private field  (basic data types like `int`) in `UserMetadata` and mutate it in public methods, it won't be possible. Also, some other data types like `sync.Mutex` or `sync.WaitGroup` should not be copied. Value receiver will restrict us from using these.

BulkBeing:
Is the content of `metadata.go` same in all packages ?
If so, one option to share code is to create a new package `internal/metadata/metadata.go` (`internal/` directory is special in Golang, only this package can import the packages under `internal/`). The re-export them:
```go
//file: pkg/mapper/metadata.go
import "github.com/numaproj/numaflow-go/internal/metadata"

type SystemMetadata = metadata.SystemMetadata
var NewSystemMetadata = metadata.NewSystemMetadata
```

pkg/mapper/metadata.go
```diff
@@ -0,0 +1,178 @@
...
+func NewSystemMetadata(d map[string]map[string][]byte) SystemMetadata {
+	if d == nil {
+		d = make(map[string]map[string][]byte)
+	}
+	return SystemMetadata{data: d}
+}
+
+// Groups returns the groups of the system metadata.
+// If there are no groups, it returns an empty slice.
+// Usage example: systemMetadata := datum.SystemMetadata()
```
BulkBeing: I think you can add code within docs using "```"  within comments.

PR #236, opened 2026-08-14
feat: introduce API to fail a message in udf/transformer

vaibhavtiwari33:
ref: https://github.com/numaproj/numaflow/pull/3570


pkg/sourcetransformer/message.go
```diff
@@ -7,6 +7,7 @@ import (
 
 var DROP = fmt.Sprintf("%U__DROP__", '\\') // U+005C__DROP__
 var NACK = fmt.Sprintf("%U__NACK__", '\\') // U+005C__NACK__
+var FAIL = fmt.Sprintf("%U__FAIL__", '\\') // U+005C__FAIL__
```
adarsh0728, line 10: nit: we can use single `var (` group here as in other files

adarsh0728 (approved):

pkg/sourcetransformer/message.go
vaibhavtiwari33: my bad, missed this. will take this up in a separate PR.

PR #238, opened 2026-09-22
fix: avoid sending second EOT during context cancellation race

### LINKED ISSUE #239: Second EOT is sent during context cancellation in batchMap
Failed flaky test: https://github.com/numaproj/numaflow-go/actions/runs/35646601978/job/106488467207?pr=229

```
=== RUN   TestService_BatchMapFn/batch_map_stream_fn_forward_msg
    service_test.go:295: BatchMapFn() got = [handshake:{sot:true} results:{keys:"client_test" value:"test1"} id:"test1" results:{keys:"client_test" value:"test2"} id:"test2" status:{eot:true} status:{eot:true}], want [handshake:{sot:true} results:{keys:"client_test" value:"test1"} id:"test1" results:{keys:"client_test" value:"test2"} id:"test2" status:{eot:true}]
--- FAIL: TestService_BatchMapFn (0.00s)
    --- FAIL: TestService_BatchMapFn/batch_map_stream_fn_forward_msg (0.00s)
```

Notice that we got two EOTs above.

vaibhavtiwari33:
Fixes: https://github.com/numaproj/numaflow-go/issues/239

In batchMap we have the following two goroutines:

https://github.com/numaproj/numaflow-go/blob/7e699ca5b8125eea1477025c200bedc21c17d0d3/pkg/batchmapper/service.go#L58-L64

First one (`receiveRequests`) receives data from the client, writes to the `datumStreamCh` channel. 
Second one (`processData`) gets data from the `datumStreamCh`, passes it on to the UDF, and returns result back to the client. At the end it always returns EOT (unconditionally)

https://github.com/numaproj/numaflow-go/blob/7e699ca5b8125eea1477025c200bedc21c17d0d3/pkg/batchmapper/service.go#L211-L224

In the concerned unit test referenced in the issue. We have the following:

https://github.com/numaproj/numaflow-go/blob/7e699ca5b8125eea1477025c200bedc21c17d0d3/pkg/batchmapper/service_test.go#L117-L139

Iteration 1: the real batch (terminated by the client's EOT request):
  - receiveRequests reads test1, test2, then the EOT request -> break, returns nil.
  - processData sends the two results, then sends EOT number 1. This is the legitimate one, and it's emitted because the client sent an actual EOT request (input item 4), not because of EOF.
  - g.Wait() -> nil, loop continues.

Iteration 2: the phantom batch (stream closed -> io.EOF):
  - receiveRequests calls Recv(), the input channel is closed -> io.EOF -> returns it, closes datumStreamCh.
  - processData runs BatchMap over the empty closed channel (no results), reaches the final select, and races the group-ctx cancellation. If it wins, it sends EOT number 2, the duplicate.


vigith (commented):
Rust won't compile it due mutable reference being shared :)

